PR #48, opened 2015-12-08
Update for most recent FreeBSD

oridb:
A couple of trivial changes: First off, FreeBSD 10.\* should get the same linker treatment as 5-9. Second, the chromium port on FreeBSD installs the binary as 'chrome'.

### DIFF
--- a/bin/9l
+++ b/bin/9l
@@ -19,7 +19,7 @@ case "$tag" in
 	5.2.*)
 		extralibs="$extralibs -lkse"
 		;;
-	[5-9].*)
+	[5-9].*|10.*)
 		extralibs="$extralibs -lpthread"
 		;;
 	esac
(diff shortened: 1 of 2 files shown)


gleydsonsoares:
% cat CONTRIBUTING.md

Plan 9 from User Space uses Gerrit instead of GitHub pull requests. See the
codereview(1) <http://swtch.com/plan9port/man/man1/codereview.html> man
page.

If you would like to use git directly instead of the codereview(1) wrapper,
that is fine too. Follow the instructions on the RE2 Contributing
<https://github.com/google/re2/wiki/Contribute> wiki page, but use
https://plan9port.googlesource.com/plan9 as the origin URL.

0intro:
This change has been merged into commit 173295929a39c25a1aa1b9c010f0f338724d0d2c. Thanks.

PR #57, opened 2016-04-26
Fix buffer overflow in libdraw/font.c

raylai:
The memmove copies too much from the old buffer. Instead of handrolling a realloc, just use realloc directly.


0intro:
Could you remove the first commit and rename the second commit's subject to something like "libdraw: fix buffer overflow in loadchar"? Thanks.

0intro:
This change has been merged into commit 94b38bdb722052838eb0d940c05995b870db4ea0. Thanks.

PR #62, opened 2016-08-09
acme: fix multiline tag dumping and loading.

lego12239:
Only the first line of multiline tag is saved and loaded.
Fix this by placing in a dump file a base64 encoded tag.

Signed-off-by: Oleg Nemanov lego12239@yandex.ru

### DIFF
--- a/src/cmd/acme/rows.c
+++ b/src/cmd/acme/rows.c
@@ -312,6 +312,23 @@ rowclean(Row *row)
 	return clean;
 }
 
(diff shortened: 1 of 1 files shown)


lego12239:
Hi, all.

plan9port acme allows to use multiline tag (lines separated by \n). It's very comfortable, because we can group, for example, search patterns on one line and edit patterns on another one (this is how my workflow looks like). But we can't save to a dump file such multiline tag.

lego12239:
New pull request is made. So i close this one.

PR #92, opened 2017-07-22
png: a pixel is 32 bits wide in RGBA, regardless of system's word size

stroucki:
Signed-off-by: Michael Stroucken <mxs@cmu.edu>


stroucki:
was merged.

PR #95, opened 2017-08-09
build: Support Arch Linux on arm64 (aarch64)

ziutek:
Arch Linux uses aarch64 as architecture/machine symbol when running on ARMv7 AArch64.

$ uname -svm
Linux #1 SMP PREEMPT Sat Jul 29 13:44:05 CEST 2017 aarch64

### DIFF
--- a/dist/buildmk
+++ b/dist/buildmk
@@ -17,5 +17,6 @@ OBJTYPE=`(uname -m -p 2>/dev/null || uname -m) | sed '
 	s;.*ppc.*;power;g;
 	s;.*alpha.*;alpha;g;
 	s;.*sun4u.*;sun4u;g;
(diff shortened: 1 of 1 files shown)


ziutek:
Of course, it should be: ARMv8 AArch64.

paulzhol:
You need to use https://plan9port-review.googlesource.com instead of github pull requests.

From https://github.com/9fans/plan9port/blob/master/CONTRIBUTING.md
>Plan 9 from User Space uses Gerrit instead of GitHub pull requests. See the codereview(1) man page.
>
>If you would like to use git directly instead of the codereview(1) wrapper, that is fine too. Follow the instructions on the RE2 Contributing wiki page, but use https://plan9port.googlesource.com/plan9 as the origin URL.

PR #98, opened 2017-08-16
Remove WSYSTYPE safeguard in mkriorules.sh

wallace-aph:
The WSYSTYPE safeguard causes rio to fail to build on macOS & previous versions, both when running the install script and running 'mk' manually. It provides no verbose, and the user is left to debug on their own. I'm not sure what the original use of the safeguard was, but I can't think of one, and it seems a little outdated.

I left mkriorules.sh (which now only contains a cat line) in the mkfile, as I figured it may have other uses.


wallace-aph:
Anything? As far as I know, this issue still exists.

PR #99, opened 2017-08-18
Add Travis CI and Coverity Scan support

0intro:
The first commit adds the Travis CI configuration and the second commit adds the Coverity Scan addon to the Travis CI configuration.

This PR is also a way to experiment how we work with PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,22 @@
 language: c
 
+env:
+  global:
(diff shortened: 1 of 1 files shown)


0intro:
The build passed successfully on Travis CI. The Coverity Scan was not executed, which is expected, because it should only be executed on the `master` branch.

PR #104, opened 2017-09-04
acme: allow file names containing spaces

rjkroege:
Permit Acme to successfully edit files whose paths contain spaces by
reversibly substituting a different rune for the offending space. The
substitute rune can be controlled with the replacespace environment
variable and defaults to '␣'.

### DIFF
--- a/src/cmd/acme/acme.c
+++ b/src/cmd/acme/acme.c
@@ -130,6 +130,12 @@ threadmain(int argc, char *argv[])
 	quotefmtinstall();
 	fmtinstall('t', timefmt);
 
+	p = getenv("replacespace");
+	replacespace = L'␣';
+	if(p != nil){
+		chartorune(&replacespace, p);
+	}
+
 	cputype = getenv("cputype");
 	objtype = getenv("objtype");
 	home = getenv("HOME");
--- a/src/cmd/acme/dat.h
+++ b/src/cmd/acme/dat.h
@@ -553,6 +553,7 @@ int			messagesize;		/* negotiated in 9P version setup */
 int			globalautoindent;
 int			dodollarsigns;
 char*		mtpt;
+Rune		replacespace;
 
 enum
 {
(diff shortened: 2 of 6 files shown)


rjkroege:
Aside: this CL was originally https://codereview.appspot.com/58800043/ and I'd forgotten about it until attempting to edit a file with a space in its name. Please take a look.

jxy:
I'm wondering if using some code in unicode PUA would be better.  Presumable standardize it within plan9port for tab, space, newline, and etc., which plan9 can't deal with.  This could solve the issue of multiline tag save/dump, too.  As long as we use PUA area consistently within plan9port (or future plan9, perhaps), it would be backward compatible and generates no conflict with normal use.

rjkroege:
I’m happy to change the character. What code point do you suggest?

lego12239:
Hi everybody.

May be a simple solution with escaping spaces with '\' would be better?

lego12239:
Like here, where i do multiline tag dump/load - https://github.com/9fans/plan9port/pull/134 .

lego12239:
I think i can convert rowdump1_tag() and rowload1_tags() from my PR to some generic util functions, that do convertion from/to escaped string form. And these functions could be used not only for multiline tag dump/load, but also for file name.

lego12239:
IMHO, this looks simpler and cleaner than solution with environment variable.
What will you say, colleagues?

blakemcbride:
Although this does, in fact, work for files and directories, it doesn't solve other space problems such as:

Font  /mnt/font/Courier New/12a/font

I like this solution but feel a more general solution would be better.

mkhl:
With this patch applied, mouse 3 on something like `:23` does not select line 23 anymore.  Plumbing something like `file:23` still opens `file` and selects line 23 there.

rsc:
I pushed a possible alternative to my plan9port fork rsc/plan9port@35c1adcc9bf6813e5a1368eb43537d88863c4378.

git fetch https://github.com/rsc/plan9port acmespaces:refs/heads/acmespaces
git checkout acmespaces

Rob and I have started talking about how to do this again and that PR
is probably a decent starting point for anyone waiting for a fix.
I'm going to close this one.

rsc:
I merged acmespaces into the master branch; no need to grab my PR.

PR #105, opened 2017-09-10
lib/moveplan9.files: add missing files

adsr:
Found with `grep -lr '#!/usr/local/plan9/bin/rc' bin/` after installing to non-default path.

Not sure if all of these belong. At least `bin/ssam` appears to.


0intro:
Thanks for noticing this.

Could you sort the file with plan9port's sort (`9 sort`) to minimize the diff and rename the subject to something like "lib/moveplan9.files: add missing files"?

PR #106, opened 2017-09-12
Fix README links and remove redundant .cvsignore files

kare:


0intro:
Could you update the subjects to something like "README: update links" and "all: remove .cvsignore files"?

The first commit looks fine. However, concerning the second commit, some .cvsignore files are still necessary in the bin sub-directories, so the directories will be existing before installing program binaries.

I think you should edit the second commit to add the following (empty) files:

 - bin/fossil/.cvsignore
 - bin/fs/.cvsignore
 - bin/venti/.cvsignore

Thanks.

mazebuhu:
I would suggest adding a `.gitkeep` file instead of the `.cvsignore `.

0intro:
Yes, .gitkeep would be fine as well.

kare:
Thank you for your feedback! I've adjusted commit's and commit messages.
I'm happy to refine commits further if need be.

kare:
For future reference: Should I do one commit per pull request? Or is two-three commits per pull request manageable even if commits are independent?

0intro:
I think it's easier to review if there is only one commit per pull request or if all the commits are related.

PR #112, opened 2017-10-03
Ported 9front paint.

TobiasHeinicke:
Paint first appeared in 9front. The 9front license is reproduced
in the related source files - the original repository is located at
https://code.9front.org/hg/plan9front

I think 9front's paint would be a great addition to the plan9port,
because it is a very lightweight solution for quick sketching.
Feel free to notify me about any wrong doings regarding this
port.


TobiasHeinicke:
Hi, thanks for your feedback, I understand your point. To avoid any API changes I added a new commit with the restructuring based on your suggestions. Hope this is what you are looking for. I might rebase this into one commit if desired. Feel free to write back if anything else is wrong with this pull request.

PR #118, opened 2017-10-16
web: teach plumbunix() to match any `$BROWSER'

gleydsonsoares:
tested with chromium and luakit on OpenBSD


gleydsonsoares:
ping...
This makes web(1) opens luakit properly.

PR #127, opened 2018-01-06
skip only the surrogate pairs

mjl-:
fontsrv wasn't rendering fontawesome icons, which uses the private use area around 0xf000.

see https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates


0intro:
Please prefix the commit subject with the program name, like "fontsrv: skip only the surrogate pairs".

0intro:
Submitted as 4ebaf18e. Thanks.

PR #131, opened 2018-01-16
Use proper type for storing ulimit values.

raylai:
rc on amd64 stores ulimit values as 32-bit int, but the limits on
OpenBSD amd64 can exceed 2^31, so "ulimit -a" shows some values as
negative. This is a problem when I want to increase my ulimit but
the hard ulimit values are printed as negative.

### DIFF
--- a/src/cmd/rc/unixcrap.c
+++ b/src/cmd/rc/unixcrap.c
@@ -58,7 +58,8 @@ eusage(void)
 void
 execulimit(void)
 {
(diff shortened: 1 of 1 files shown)


0intro:
Could you rename your commit message to something like "rc: use proper type for storing ulimit values"? Thanks.

raylai:
Done.

0intro:
Merged. Thanks.

PR #133, opened 2018-01-20
Get network stats in a portable manner on FreeBSD

bakul:
as the old grody way doesn't work any more on FreeBSD-10 and later.


0intro:
Could you change the commit message to something like "auxstats: get network stats in a portable manner on FreeBSD". Thanks.

bakul:
git diff will show that a file in auxstats changed and the commit message describes what the change does. Why force another commit?

bakul:
Thanks.

0intro:
Merged. Thanks.

PR #134, opened 2018-01-21
acme: fix multiline tag dumping and loading.

lego12239:
Only the first line of multiline tag is saved and loaded.
Fix this by doing:
- replace of '\n' with '\\n' on file dump
- replace of '\\n' with '\n' on file load

Introduce acme dump format version mark(the first line of
acme dump file) in format "ACMEDUMP"VERSION_NUMBER.
Current version, for example, has version mark "ACMEDUMP1".

Support the loading of old acme dump format.

Signed-off-by: Oleg Nemanov <lego12239@yandex.ru>

### DIFF
--- a/src/cmd/acme/rows.c
+++ b/src/cmd/acme/rows.c
@@ -313,8 +313,8 @@ rowclean(Row *row)
 	return clean;
 }
 
-void
-rowdump(Row *row, char *file)
+static void
+rowdump0(Row *row, char *file)
 {
 	int i, j, fd, m, n, dumped;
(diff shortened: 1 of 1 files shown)


lego12239:
Reworked old patch with taking into account Gleydson Soares and Russ Cox comments.

lego12239:
plan9port acme allows to use multiline tag (lines separated by \n). It's very comfortable, because we can group, for example, search patterns on one line and edit patterns on another one (this is how my workflow looks like). But we can't save to a dump file such multiline tag.

PR #139, opened 2018-01-31
mount: Check current osxfuse kext location

mkhl:
Current versions of osxfuse ship with multiple versions of its kernel
extension (kext) for differend versions of macOS.

Running mount(1) on macOS with a current version of osxfuse fails with
`don't know how to mount (no fuse)' since it fails to find the kext.
Running 9pfuse(4) directly works fine.

This change adds a check to mount(1) that determines:
1) which version of macOS we're running on
2) if there is an osxfuse kext available for this version of macOS


0intro:
Submitted as 72fc31ac. Thanks.

PR #141, opened 2018-02-01
git: Ignore files created for astro(1) and scat(1)

mkhl:
To use astro(1) and scat(1) one has to create sky/here and download
various catalogue files as detailed in sky/README.
This change marks those files as ignored by git so they don't clutter
its status messages.


0intro:
Submitted as 3473f4e5. Thanks.

PR #145, opened 2018-02-06
9term: change behavior of $SHELL to split it by spaces

lufia:
I'm not sure that *$SHELL* could contain arguments.

But, in case of launching acme directly from Dock in macOS, like as **/mac/9term.app**, `rc` that is started from `win` can't load *$home/lib/profile* before showing a prompt.
To solve it, I think set `SHELL='rc -l'` is useful.

### DIFF
--- a/src/cmd/9term/rcstart.c
+++ b/src/cmd/9term/rcstart.c
@@ -34,18 +34,20 @@ int
 rcstart(int argc, char **argv, int *pfd, int *tfd)
 {
 	int fd[2], i, pid;
-	char *cmd, *xargv[3];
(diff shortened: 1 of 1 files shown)


lufia:
Thank you for a reply.

I can run *$home/lib/profile* with a script that just do `rc -l`, but this work around has a dependency on a external resource of plan9port.
Therefore, in case of publishing scripts or macOS app bundles, the distribution should include this simple script in all app bundles. And also scripts or app bundles should be careful to set correct *$PATH*.

rsc:
I made 9term -l pass -l to the shell, and I made 9term.app run 9term -l.

PR #147, opened 2018-02-28
cmd/9term: Set TERM_PROGRAM to termprog

mkhl:
TERM_PROGRAM is the customary way to identify which kind of terminal
emulator program one uses on macOS.
This change sets TERM_PROGRAM to termprog since both variables are used
for the same purpose.

For some examples, see:
https://apple.stackexchange.com/questions/42714/what-is-the-term-program-and-term-program-version-environment-variables-used-fo
https://github.com/jwilm/alacritty/issues/781
https://github.com/justinmk/vim-gtfo/issues/3
https://github.com/neovim/neovim/blob/f050aaabbb31b58ebac519a6f7014610f2b119f0/src/nvim/tui/tui.c#L206-L211


0intro:
Submitted. Thanks.

PR #150, opened 2018-03-15
9pfuse: retries read(3) upon EINTR

jxy:
read(3) sometimes errors with EINTR on macOS over slow connections.
9pfuse(1) now retries read(3) instead of sysfatal(3)ing.


0intro:
Submitted. Thanks.

PR #153, opened 2018-03-20
acme: allow multi-line window tag dump/load encoded with unicode PUA

jxy:
All newline (0X0A) runes are replaced by 0XF00A when dumping, and
swap back when loading.  Due to utf-8 encoding, the ondisk byte
sequence is 0XEF808A.

### DIFF
--- a/src/cmd/acme/rows.c
+++ b/src/cmd/acme/rows.c
@@ -22,6 +22,12 @@ static Rune Lcolhdr[] = {
 	0
 };
 
+/*
+ * Use a character from the Unicode Private Use Area to escape newlines
+ * within multi-line tags.
+ */
+#define PMASK(x) (0XF000^x)
+
 void
 rowinit(Row *row, Rectangle r)
 {
@@ -433,10 +439,10 @@ rowdump(Row *row, char *file)
 			Bwrite(b, buf, strlen(buf));
 			m = min(RBUFSIZE, w->tag.file->b.nc);
 			bufread(&w->tag.file->b, 0, r, m);
-			n = 0;
-			while(n<m && r[n]!='\n')
-				n++;
-			Bprint(b, "%.*S\n", n, r);
+			for(n=0; n<m; n++)
+				if(r[n] == '\n')
+					r[n] = PMASK('\n');
+			Bprint(b, "%.*S\n", m, r);
 			if(dumped){
 				q0 = 0;
 				q1 = t->file->b.nc;
@@ -731,6 +737,9 @@ rowload(Row *row, char *file, int initing)
 			if(r[n] == '|')
 				break;
 		wincleartag(w);
+		for(i=n; i<nr; i++)
+			if(r[i] == PMASK('\n'))
+				r[i] = '\n';
 		textinsert(&w->tag, w->tag.file->b.nc, r+n+1, nr-(n+1), TRUE);
 		if(ndumped >= 0){
 			/* simplest thing is to put it in a file and load that */


0intro:
It looks good to me, but I'd like @rsc or @robpike to review this change.

robpike:
I have no idea what this is about or what it fixes.

jxy:
When acme `Dump`s, multi-line window tags are cut off before `\n`, and the `acme.dump` only contains the first line of a window tag.  This patch replaces all `\n` in window tags with a rune (0xf00a) in unicode PUA region, when `Dump`, and converts them back to `\n` when `Load`.

This is similar to #135, but I feel this is much simpler, and have almost perfect backward compatibility.

robpike:
OK, I understand now. The PR description could have been much more informative.

I'm OK with merging this in preference the #135

0intro:
@jxy Maybe you could add the description of the issue to the commit message, so it would become clearer what is the problem being fixed when reading the Git history.

lego12239:
hi.

@jxy Our company, for example, use PUA for own purposes. And 0xF00A can occur in files are edited by me. And what do i have i this case in dump file?

Thus i think that doing replacing instead of escaping is fundamentally wrong in this case. This approach is simple. Yes. But this can cause an unexpected behaviour from users point of view in the future.

About backward compatibility. My patch is fully backward compatible. Moreover it adds such important thing as dump format version mark and actually supply something like a framework for future dump formats code with backword compatibility with every previous formats.

jxy:
PUA is supposed to be used for private purposes of any group.

Only ``0XF00A`` in window tag is converted to ``0X0A`` upon ``Load``.  I guess it can cause headaches when you ``Load`` a dump file and see too many new lines in your window tag.  I guess people actively editing files that have symbols such as https://fontawesome.com/icons/th can have problems too.

Actually, PUA is already used in ``keyboard.h``. https://github.com/9fans/plan9port/blob/03a8ec739af17bc4ba0a2e18ea59b33671c34f2b/include/keyboard.h#L18-L43
If you type ``Alt X f 0 1 1`` you can see your cursor moved to the left.

I did choose simplicity over consistency.

lego12239:
> I did choose simplicity over consistency.

jxy, i understand your point of view. But it sounds strange for me :-). IMHO, consistency in task solving is go first. And only after it comes simplicity. That is, task must be solved and we should do this as simple as possible. Not vice versa.

Look, our world full of limitations and exceptions. We must remember how many bytes in integer on target architecture, what max file size for given fs, how many bytes can be send in one packet for a given protocol, what flowers to buy to makes your wife happy, etc. Why do we need to add yet another exception to this very big list in our brain if we can simply avoid it?

About simplicity. IMHO, what really will make code simpler is refactoring of at least rowdump() & rowload(). rowload() is about 300 lines long! And my patch is simple as simple the current rowload() and rowdump() code, because all it actually do is redirecting to current rowload() and rowdump() or new rowload() and rowdump() according to dump format version.

jxy:
@lego12239 Some part of me actually want your solution, sans the dump file versioning.  The actual breakage of interpreting `\n` is really minimal to be of any concern.  If I were doing release cycles of p9p, I would just introduce this change in a major version update without hesitation.  If there were enough interests.

lego12239:
@jxy Why don't you like dump file versioning :-)?

src/cmd/acme/rows.c
```diff
@@ -22,6 +22,9 @@ static Rune Lcolhdr[] = {
 	0
 };
 
+/* Use unicode PUA to (un)mask ascii */
+#define PMASK(x) (0XF000^x)
```
robpike: The comment is poor.
/*
 * Use a character from the Unicode Private Use Area to escape newlines
 * within multi-line tags.
 */

src/cmd/acme/rows.c
```diff
@@ -433,10 +436,10 @@ rowdump(Row *row, char *file)
 			Bwrite(b, buf, strlen(buf));
 			m = min(RBUFSIZE, w->tag.file->b.nc);
 			bufread(&w->tag.file->b, 0, r, m);
-			n = 0;
-			while(n<m && r[n]!='\n')
-				n++;
-			Bprint(b, "%.*S\n", n, r);
+			for(n=0;n<m;n++)
```
robpike: Please look at the style in the surrounding code. There should be spaces after the semicolons in the for and around the operator (==) in the if.

jxy:
I hope it is now up to standard.

@lego12239 Just more things to maintain.  Besides, such a small breaking change is totally justifiable in a new release.

robpike:
Seems fine to me now.

lego12239:
@jxy, hm. More to maintain - may be(but not many!). But we get 2 important features:

1. smooth and transparent migration from old dump format to new one for all users;
   all we need for this just to keep previous dump format load code(load old dump format, and save new one); all earlier formats code can be safely removed from sources;
2. if we see very old dump format, we don't try to load it and got something strange - just show a message "unknown dump format".

lego12239:
@jxy, hm. More to maintain - may be(but not many!). But we get 2 important features:

1. smooth and transparent migration from old dump format to new one for all users;
   all we need for this just to keep previous dump format load code(load old dump format, and save new one); all earlier formats code can be safely removed from sources;
2. if we see very old dump format, we don't try to load it and got something strange - just show a message "unknown dump format".

mjibson:
I tried out this patch. It appeared to correctly write the dump file (the extra lines showed up in it) but did not restore them on startup.

robpike:
I'm starting to think there is a better approach, so I am withdrawing my support for this change.

robpike:
I will when I have a working example. I've been making progress but nothing to report yet.

PR #163, opened 2018-03-31
TLS fix for upas/smtp

ethhics:
This PR fixes the TLS functionality in `upas/smtp`. With these fixes, I am able to connect to TLS enabled email servers like Gmail and send messages using Acme Mail and `mail`.

Both `upas/nfs` and `upas/smtp` used `tlsClient()` in the original plan9 code. In the nfs function `imapdial()`, the call to tlsClient was replaced with `stunnel3` on ported systems. This PR simply replaces the tlsClient call in smtp with stunnel as well.

Because tlsClient was called differently in the two (nfs called the userland function, while smtp used the library function), I had to patch a copy of imapdial called `smtpdial()` into the smtp code. smtpdial then replaces the dial call in the mxdial function.

I am very open to suggestions on how to make this solution more elegant!

### DIFF
--- a/src/cmd/upas/smtp/mxdial.c
+++ b/src/cmd/upas/smtp/mxdial.c
@@ -2,6 +2,7 @@
 #include <ndb.h>
 #include "smtp.h"	/* to publish dial_string_parse */
 #include <ip.h>
+#include <thread.h>
 
 enum
 {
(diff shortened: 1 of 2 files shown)


0intro:
At first glance, it looks fine, but I'd like to try on my side.

Could you rename you commit subject to something like "upas/smtp: fix TLS" and add the description from the PR to the commit message? Thanks.

josuah:
Mind that stunnel can not be built with LibreSSL [by choice of its autho](
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=198997#c2).  This pulls a dependency to OpenSSL into upas.

Using `openssl s_client` would kill one dependency (as it will be present if stunnel is, and libressl keeps the same command name and arguments).

Although maybe there are reasons to use stunnel in particular.

Just for fun: [`brssl client`](https://bearssl.org/gitweb/?p=BearSSL;a=blob;f=tools/client.c;h=3d13510befc09fe2b50abe457a5bc2d0df51a30a;hb=HEAD#l459) can be used with inetd to build a local proxy, then it gets possible to have one local TCP port per TLS connection to open, all of that in one inetd or alike.

josuah:
I will come back with more on this if I get it working.

josuah:
Thank you for enlightening me up on this.

PR #167, opened 2018-05-10
add ttf2subf for convert ttf font to plan9 (subf) format

ignusius:
Convert your lovely fonts and use it:

> ttf2subf -s 14 -f example.ttf -n example -m antialias
> acme -f ~/you/path/example/example.14.font


ignusius:
I and more people use only acme from plan9port ,fontsrv is redundantly.

P.S. It's my opinion.

0intro:
If you don't want to depend on plan9port too much, I think the best solution would be to port Andrey Mirtchovski's [ttf2subf](http://mirtchovski.com/p9/freetype/) back to a POSIX environment.

Also, it would allow to depend on the (recent) FreeType library provided by your operating system.

PR #169, opened 2018-05-17
Correctly quote IMAP LOGIN command arguments

cac04:
According to RFC 3501 the arguments to the LOGIN command should be quoted strings (or length prefixed string literals). Without quoting, authentication to some IMAP servers (e.g. Dovecot) will fail.

### DIFF
--- a/src/cmd/upas/nfs/imap.c
+++ b/src/cmd/upas/nfs/imap.c
@@ -214,7 +214,7 @@ imaplogin(Imap *z)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


0intro:
It seems good to me. Could you prefix the name of the program to the commit subject, like "upas/nfs: correctly quote IMAP LOGIN command arguments". Thanks.

cac04:
Sure, I've done that now.

0intro:
Sorry, I forgot about it. I'll submit your change.

PR #170, opened 2018-05-17
fontsrv: omit box-drawing characters from line struts on macOS

### LINKED ISSUE #162: fontsrv: box-drawing characters as representative text make line height too large for some fonts
In recent commit 96dc233, box-drawing characters were [added into the
`lines` array][1] of representative texts used by fontsrv to compute
the line height of a served font on macOS.

Unfortunately, this change artificially increases the resulting line
height to an uncomfortably high value in some cases. For instance, if
the box characters are missing in the font, their versions from the
whatever substitution font macOS falls back to are used instead, and
those replacement glyphs may mismatch the height of the original font
in question, being too tall for it.

Below are screenshots of Acme windows running with the same settings
against fontsrv without and with box-drawing characters in the
representative texts array (i.e., first with line
[`src/cmd/fontsrv/osx.c:107`][1] commented out and then with it being
left intact).

&nbsp;&nbsp;

Perhaps, the isue might also apply to those fonts that do have
box-drawing characters in them, but in which the corresponding glyphs
are significantly taller than glyphs for other characters.

[1]: https://github.com/9fans/plan9port/blob/03a8ec739af17bc4ba0a2e18ea59b33671c34f2b/src/cmd/fontsrv/osx.c#L107

iburago:
For some fonts, using box-drawing characters in the representative
text for computing the line height results in it being uncomfortably
high. Replace them with accented capitals and tall lower-case letters
which lead to a more conservative increase in the line height.

Fixes #162.

### DIFF
--- a/src/cmd/fontsrv/osx.c
+++ b/src/cmd/fontsrv/osx.c
@@ -104,7 +104,7 @@ static char *lines[] = {
 	"ABCDEFGHIJKLMNOPQRSTUVWXYZ",
 	"abcdefghijklmnopqrstuvwxyz",
 	"g",
-	"┌┬┐├┼┤└┴┘│─",
+	"ÁĂÇÂÄĊÀČĀĄÅÃĥľƒ",
 	"ὕαλον ϕαγεῖν δύναμαι· τοῦτο οὔ με βλάπτει.",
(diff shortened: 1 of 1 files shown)


0intro:
@jxy What do you think about this change?

Issue #125, opened 2017-11-17
acme crashes when displaying some characters


rogpeppe:
I'm using the Go font like this:

    fontsrv &
    $PLAN9/bin/acme -f '/mnt/font/Go Regular/13a/font' -a $*

when I view $PLAN9/lib/keyboard (and other files with unicode or
binary characters), acme crashes with:

    acme: end of string in frcharofpt: No such file or directory

a) it should be more resilient to bad fonts
b) it should at least do a Dump before exiting, so work is not lost.

rsc:
48da9bd is supposed to fix this on x11, and I can't reproduce on macOS, so closing.

PR #173, opened 2018-06-22
fontsrv: copy some fixes from OS X to X11

fhs:
* Avoid allocating empty images by adding 1 to width/height. This was
  crashing fontsrv. The total width of the subfont image can be zero
  even if the characters are present in the font. For example, all the
  characters in x0300.bit (part of "Combining Diacritical Marks" Unicode
  block) have zero width.
* Enable all Unicode ranges. We need to enable U+0000, otherwise libdraw
  complains with: "stringwidth: bad character set for rune 0x0000 in ..."
* Use the same fallback glyph (pjw face) as OS X. This also fixes a bug
  where advance was set to the total width of subfont instead of the
  character.

Update #125 (most likely fixes the crash if in X11)

CC @paulzhol

### DIFF
--- a/src/cmd/fontsrv/x11.c
+++ b/src/cmd/fontsrv/x11.c
@@ -66,19 +66,16 @@ load(XFont *f)
 		return;
 
 	e = FT_New_Face(lib, f->fontfile, f->index, &face);
-
 	if(e){
 		fprint(2, "load failed for %s (%s) index:%d\n", f->name, f->fontfile, f->index);
 		return;
 	}
-
 	if(!FT_IS_SCALABLE(face)) {
 		fprint(2, "%s is a non scalable font, skipping\n", f->name);
 		FT_Done_Face(face);
-	    f->loaded = 1;
+		f->loaded = 1;
 		return;
 	}
-
 	f->unit = face->units_per_EM;
 	f->height = (int)((face->ascender - face->descender) * 1.2);
 	f->originy = face->descender; // bbox.yMin (or descender)  is negative, becase the baseline is y-coord 0
@@ -96,7 +93,11 @@ load(XFont *f)
 			f->nrange++;
 		}
 	}
-
+	// libdraw expects U+0000 to be present
+	if(!f->range[0]) {
+		f->range[0] = 1;
+		f->nrange++;
+	}
 	FT_Done_Face(face);
 	f->loaded = 1;
 }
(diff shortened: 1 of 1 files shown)


src/cmd/fontsrv/x11.c
```diff
@@ -59,44 +59,31 @@ load(XFont *f)
...
-		int idx = charcode/SubfontSize;
-
-		if(charcode > 0xffff)
-			break;
-
-		if(!f->range[idx]) {
-			f->range[idx] = 1;
-			f->nrange++;
-		}
+	// enable all Unicode ranges
```
paulzhol: I'm not sure I understand this change, the original code would scan the glyphs in the font and mark the present ones. Now you mark all as present, even if they are not?

https://www.freetype.org/freetype2/docs/reference/ft2-base_interface.html#FT_Get_Next_Char

fhs: `libdraw` expects U+0000 to be present. Initially, I had the code make sure U+0000 was always present, but then I saw OS X enables all ranges, so I did the same. Since we now make sure empty subfonts don't cause any problem, it doesn't hurt to enable everything.

paulzhol: I think the OSX code was changed because it used `CGFontGetGlyphsForUnichars` which is not available anymore(?), I don't feel very strongly about it so +1 and thanks again!

jxy: OSX enabling all the ranges is because it fills in the missing glyphs with glyphs from a set of default fonts.  Do you think you can make freetype do it?  Or do we need fontconfig for backup font families?

jxy:
While you are at it, is it possible to remove the spaces in the font names?  Or just replace the spaces in the font names with underscores?

fhs:
I've changed it back to scanning for available characters, instead of enabling all Unicode ranges. Also, I replaced FT_Load_Char with FT_Get_Char_Index+FT_Load_Glyph. FT_Load_Char will actually return a rectangle glyph for missing characters, so you end up with that instead of pjw face. This makes it consistent so that you always see pjw for missing characters.

fhs:
> While you are at it, is it possible to remove the spaces in the font names? Or just replace the spaces in the font names with underscores?

I'm not sure why you want to do that. It has potential to break things.

jxy:
>> While you are at it, is it possible to remove the spaces in the font names? Or just replace the spaces in the font names with underscores?

> I'm not sure why you want to do that. It has potential to break things.

So that I can change font within acme.

I just checked, OSX uses ``kCTFontNameAttribute``, which gets the PostScript name.

Perhaps, here

https://github.com/9fans/plan9port/blob/03a8ec739af17bc4ba0a2e18ea59b33671c34f2b/src/cmd/fontsrv/x11.c#L43

instead of ``FC_FULLNAME``, we should use ``FC_POSTSCRIPT_NAME``, which is guaranteed to contain no spaces, and makes it the same as OSX does now.

jxy:
I made the postscript name PR #174

fhs:
ping? I think this has been tested enough, so it can be merged?

0intro:
It seems good to me. Thanks for your work.

PR #176, opened 2018-07-23
Fixing tar options for mkfile in unix directory

mek:
Was playing around and attempted mk tgz in plan9front/unix and it failed (create a '-' tar archive). Per tar man file, this version of tar should use tar c <name> to generate an archive from standard out.

This is just a simple patch to make that change.

### DIFF
--- a/unix/mkfile
+++ b/unix/mkfile
@@ -38,15 +38,15 @@ test-%:V:
 
 lib%.tgz:V:
 	mk new-$stem
-	tar cf - lib$stem | gzip > $target
+	tar c lib$stem | gzip > $target
 
 libregexp9.tgz:V:
 	mk new-regexp
-	tar cf - libregexp | gzip >$target
+	tar c libregexp | gzip >$target
 
 mk.tgz:V:
 	mk new-mk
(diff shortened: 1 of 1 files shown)


0intro:
What version of tar are you using? Both "tar cf -" and "tar c" syntaxes work with either GNU tar and plan9port tar.

jacobvosmaer:
@0intro looking at this diff, I think the problem is that `cd unix && 9 mk` would invoke plan 9 tar, which then fails (because `9 tar cf -` creates a file literally called `-`). Perhaps unix/mkfile is not meant to run with plan 9 tar?

jacobvosmaer:
I'm just a drive-by commenter on this PR. @mek I think [this question](https://github.com/9fans/plan9port/pull/176#issuecomment-438541494) is for you.

rsc:
I was worried about BSD tar but at least the BSD tar on my Mac is OK with 'tar c'. Will apply.

jxy:
@rsc Your worry was correct.  This breaks on FreeBSD.

jxy:
A compromise would be `tar cf /dev/stdout`.

PR #181, opened 2018-09-21
Add '@' to list of characters expected to be part of file/directory name.

gcapell:
Mainly because Go's pkg/mod directory is full of directories
named pkg@version.


rsc:
Is this needed? I feel like I was able to just update my plumb rules. The implication here seems to be that you're not using the plumber?

gcapell:
I hadn't realised plumb rules could help with that. Duh. Thanks, and ignore
this PR.

rsc:
Fixed in 82abcd6fd693.

PR #182, opened 2018-09-21
plumber: Make `@` valid in all paths

mkhl:
Closes #181


rsc:
Fixed in 82abcd6fd693.

PR #186, opened 2018-10-01
fix segfault in grep -e if no argument was provided

henesy:
Previously if you performed `grep -e` with no arguments you'd get a segfault.

After adding a comparison for nil there is no segfault with a `grep -e`  and you get usage as would be expected.

I don't like adding a variable for this, but I didn't see a better way immediately. Thoughts?

### DIFF
--- a/src/cmd/grep/main.c
+++ b/src/cmd/grep/main.c
@@ -13,6 +13,7 @@ void
 main(int argc, char *argv[])
 {
 	int i, status;
+	char* arge;
 
 	ARGBEGIN {
 	default:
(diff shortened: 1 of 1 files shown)


lfdebrux:
Did you try using `EARGF(usage)`?

https://9fans.github.io/plan9port/man/man3/arg.html

henesy:
Good catch! I totally missed that on my first pass, my b. Latest commit uses EARGF() and it's much cleaner.

0intro:
Thanks for your PR. The fix is fine, but this issue has already been [fixed](https://github.com/0intro/plan9/commit/dc33830ed7115a0c89ebb0b3ab40b05c8eb26cf8#diff-24f380131cc0d72ec243508fa75b8cfb) in Plan 9 (2004-08-08) using EARGF and I think it's preferable to synchronize code between plan9port and Plan 9.

I'll submit a change to synchronize plan9port's grep with Plan 9's.

Issue #185, opened 2018-09-26
devdraw: After MacOS Mojave update Acme window remains gray


palmamartin:
Hello,
sorry for the noise but after upgrading to MacOS Mojave and rebuilding plan9port my Acme window remains gray. Windows resizing didn't help either...  See:

<img width="968" alt="screenshot 2018-09-26 at 09 45 39" src="https://user-images.githubusercontent.com/381/46065030-4fdaba80-c171-11e8-93c0-748b3ab63e41.png">

Any suggestions?

mazebuhu:
Find here some logs with backtrace of devdraw when starting acme: [https://gist.github.com/mazebuhu/5ae3cf2c3d5144ff46df571a6bd949da](https://gist.github.com/mazebuhu/5ae3cf2c3d5144ff46df571a6bd949da)

mjl-:
can confirm. just did a fresh plan9port build from a fresh git clone on mojave. devdraw only shows a gray screen.

jxy:
Yes.  I have the same problem.  Maybe we need to use `-drawRect` and remove the custom scheduling of refresh.

mjl-:
true, the problem is with binaries compiled on mojave. before i rebuilt, i had binaries from the previous macos release, and those still worked on mojave.

so don't yet recompile after updating to mojave. (:

btw, i'm using lato as acme font, and thought was it looked worse. which may have triggered me to pull latest plan9port and recompile.

mazebuhu:
Yes binaries build pre Mojave will work. The issue occurs only after a
rebuild or a build from scratch on Mojave.

jxy:
I copied `devdraw` only from the time machine and everything else works.  @mazebuhu perhaps you can change the title to mention `devdraw` instead.

iburago:
@kare @jxy Would you mind sharing a copy of the `devdraw` binary
built on High Sierra for those of us who do not have it and need to do
a fresh install on Mojave (as opposed to rebuilding after updating to it)?

mazebuhu:
@iburago find mine here: https://drive.google.com/file/d/1L9gig6UdGErLeoNb7BhFMQLnhczMGDh3/view?usp=sharing

iburago:
@mazebuhu Thank you!

jacobvosmaer:
[...]

PR #188, opened 2018-10-05
WIP try to fix devdraw in macOS 10.14

jacobvosmaer:
Messing around, trying to get traction on https://github.com/9fans/plan9port/issues/185

This is work in progress and may not go anywhere. I am not an experienced macOS developer.

## How to try out this executable

- Clone https://github.com/jacobvosmaer/plan9port.git
- checkout `devdraw-macos-1014`
- in that checkout: `cd src/cmd/devdraw && 9 mk`

Now you have `src/cmd/devdraw/o.devdraw`. If you rename this file to `devdraw` you can drop it in your `plan9port/bin`. Make sure to keep a copy of your old devdraw! I recommend just making a copy or zip of your entire plan9port directory so you can get it back to a working state.


jacobvosmaer:
I don't really understand the architecture of devdraw, nor do I understand how one is supposed to use AppKit (NSWindow, NSView and friends). But it looks like devdraw uses AppKit in an "unintended" way and as if macOS 10.14 broke this use case. I'm trying to bring the code in line with what I think Apple wants us to do.

This may be way off base. I struggle, a lot, to understand "what Apple wants us to do", and I have no reason why devdraw does things the way it does.

jacobvosmaer:
But the good news is that I got it to draw the acme interface and I can click around in it. The moment I resize it's toast, though.

jacobvosmaer:
I should also add I'm working on a non-retina macbook. Wouldn't be surprised if retina support is broken at the current commit.

jacobvosmaer:
At this point the thing sort of works (on my non-retina machine anyway). The window doesn't update when I resize it, but eventually it gets redrawn. I can go to and from full-screen. Scrolling seems slower than it used to be, which is probably my fault.

I don't know if my current changes are going in a direction that makes sense. There seems to be an Apple way of doing things, and then devdraw intentionally does not do things the Apple way. My changes take devdraw closer to the Apple way. Is that going to work?

One example of this: apparently, certain AppKit method calls from within the drawRect method automatically use the right graphics context, and all sorts of things are just taken care of by AppKit for you. For example, rounding the corners of the window. But for some reason devdraw is manipulating the graphics context from outside drawRect, and creating its own rounded corners effect -- manually! This also seems to be the place where things break down in macOS 10.14. The code that tries to draw, from outside drawRect, is failing to get a graphics context, making all its drawing calls fail at runtime. All I'm seeing in macOS 10.14 release notes suggests "don't grab the graphics context, use drawRect". So I think this might just be what we have to do.

Then there is the question of devdraw. Is it multi-threaded? How do the devdraw events reach the AppKit (Cocoa) code? What sort of synchronization are we using? There is weird stuff going on with drawRect spinning and waiting for the value of an int to flip from 0 to 1. Is the assumption that reading/writing an int is atomic? I don't see locks anywhere. 

So many questions ;)

jacobvosmaer:
I wrote above that the current version of this branch doesn't behave right when resizing. I forgot that acme doesn't update while resizing anyway. So that sounds pretty much normal.

What is not normal is that after resizing, the window isn't immediately redrawn. But that sounds like a much easier problem to solve.

jacobvosmaer:
I think I'm starting to understand the design. There are two threads, one for AppKit (Cocoa) and one for the devdraw server. 

The AppKit thread manages an NSWindow, a NSView subclass and NSImage. The NSImage is backed by a piece of memory that devdraw writes to.

The devdraw thread draws in this piece of memory without any sort of locking. Occasionally, devdraw will call `_flushmemscreen`. This method crosses the boundary into AppKit, and causes a method call on the AppKit thread. This thread reads from the bitmap and writes it, or a section of it, into the NSView. I think the reason this is OK is that devdraw blocks while this is happening.

I think this is how the existing code was working around the problem of shared access to the image data. The problem is that it looks like macOS 10.14 is making it harder, or impossible, to do this trick of updating the NSView outside drawRect.

mjl-:
I tried your branch on my Mojave install just now. No more grey screen, the acme UI draws both on regular and retina screens, resize also seems to work, nice! I do see "refresh flashing", e.g. when changing cursor location through a click in acme. Double click in acme often doesn't work because of this. You're probably aware of this (this being WIP), but if you haven't seen this let me know.

Thanks for working on this! I don't have enough devdraw-knowledge and time to be of more help at the moment, but I can do more testing.

jacobvosmaer:
Thanks @mjl! I've been busy, I made a lot of changes and I think it works better now.

jacobvosmaer:
I found a silly way to benchmark my new rendering code: just create a program or script that blurts a lot of data to stdout, and use `time` on that. By this metric the new renderer is now on par or slightly faster than the old one (on my machine, on MacOS 10.14). That suggests it's "good enough" I think.

jacobvosmaer:
Closing in favor of #191 which is this, minus the 10.5 deletions, and with the commits squashed.

PR #190, opened 2018-10-08
devdraw: remove os x 10.5 compatibility code

jacobvosmaer:
In `src/cmd/devdraw` we still have code that targets OS X 10.5 (Leopard) and older. The last release of OS X 10.5 was 10.5.8 in 2009, according to Wikipedia. Shall we delete this old code?


rsc:
Merged.

PR #191, opened 2018-10-08
devdraw: rewrite cocoa renderer

jacobvosmaer:
The old renderer did a lot of work outside the Cocoa drawRect method.
In macOS 10.14 this stopped working (see
https://github.com/9fans/plan9port/issues/185). The best information
I could find on how to remedy this was the suggestion to "just use
drawRect".

This ended up as a substantial rewrite of the rendering code. It
looks to me like we did not take a performance hit in doing so. I
think the code is simpler now because it lets Cocoa take care of
more things for us.

I have manually tested this code on macOS 10.13 (retina) and 10.14
(non-retina). I don't know how it behaves on 10.12 and before.

To clean up while I'm here I have removed ifdef's in cocoa-screen.m
for 10.6 support. The last 10.6 release was in 2009.

### DIFF
--- a/src/cmd/devdraw/cocoa-screen.m
+++ b/src/cmd/devdraw/cocoa-screen.m
@@ -142,9 +142,13 @@ struct
 	NSCursor		*cursor;
 	CGFloat		topointscale;
 	CGFloat		topixelscale;
-	Memimage	*imgcocoa;
+
+	/*
+	 * The following fields are guarded by imgcocoalock
+	 */
 	QLock		imgcocoalock;
-	int		needsinitimg;
+	Memimage	*imgcocoa;
+	int		needsflushmem;
 
 	/*
 	 * imgdevdraw is allocated on the main thread but all other access is on the
@@ -205,7 +209,10 @@ static NSRect dilate(NSRect);
 {
 	getmousepos();
 	sendmouse();
-	win.needsinitimg = 1;
+
+	qlock(&win.imgcocoalock);
+	win.needsflushmem = 1;
+	qunlock(&win.imgcocoalock);
 }
 - (void)windowWillStartLiveResize:(id)arg
 {
@@ -438,7 +445,6 @@ makewin(char *s)
 	}
 	win.isofs = 0;
 	win.imgcocoa = nil;
-	win.needsinitimg = 1;
 	win.content = [contentview new];
 	[WIN setContentView:win.content];
 
@@ -497,6 +503,7 @@ initimg(void)
 		panic("allocmemimage: %r");
 	if(win.imgcocoa->data == nil)
 		panic("win.imgcocoa->data == nil");
+	win.needsflushmem = 1;
 	qunlock(&win.imgcocoalock);
 
 	ptsize = winsizepoints();
@@ -510,7 +517,6 @@ initimg(void)
 	// http://en.wikipedia.org/wiki/List_of_displays_by_pixel_density#Apple
 	displaydpi = win.topixelscale * 110;
 
-	win.needsinitimg = 0;
 	return win.imgdevdraw;
 }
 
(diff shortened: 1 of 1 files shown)


jacobvosmaer:
This closes #185 and replace #188.

jacobvosmaer:
I am not an experienced plan9port contributor, nor do I have graphics programming experience, nor non-trivial macOS development experience. So if something I do in this code looks funny, please let me know.

jacobvosmaer:
Maybe "substantial rewrite of the rendering code" is a bit of an overstatement but I had to change more than I expected. I ended up deleting a fair bit of code that I found hard to understand, and of which I can only think there must have been a reason it looked that way. I do have a good feeling about the code that came in its place -- with the proviso that it's guesswork that is not backed by experience.

RobbinJohnson:
About to check this out on a couple of different 10.14 systems (one retina, one non-retina).  Your change addresses two separate issues: the subject devdraw issue in Mojave and deleting support for 10.7 and earlier.  Would you consider separating the two?  I have an old G5 iMac which cannot run anything later than 10.5.8 and I run plan9ports on it (I mean, what else can you do with a 13 year old Mac?)

jacobvosmaer:
@RobbinJohnson OS X 10.5 uses Carbon, not Cocoa. I am not touching Carbon support in this PR (but I opened !190 to have it removed).

It's not up to me (I am not a maintainer of this project) but I think keeping support for so many old OS's is a bad idea. The complexity adds up, making maintenance harder and making it harder to contribute. It's not good for the long term health of this project.

But that's just my opinion. Let's see what the maintainers say about the OS X 10.7 ifdefs when this gets reviewed.

jacobvosmaer:
I may have gotten it wrong in my commit message. This commit is only removing support for OS X 10.6. It's keeping the code that is for 10.7 and up. As mentioned above, 10.5 code resides in another file.

RobbinJohnson:
Thank you for the clarification!

sevan:
> I may have gotten it wrong in my commit message. This commit is only removing support for OS X 10.6. It's keeping the code that is for 10.7 and up. As mentioned above, 10.5 code resides in another file.

But you are removing 10.5 support in #190

jacobvosmaer:
I have clarified the MR description and force-pushed a new commit that matches it.

prodhe:
Just wanted to say thank you for this one. I have no idea how it affects the overall system and older versions, but It Works On My Machine (TM), MBA 11" 10.14.

jacobvosmaer:
Thanks @prodhe 

While using this myself I encountered an (aesthetic) bug where there could be a drawRect call before devdraw had received any draw operations from its client (e.g. Acme). This would lead to a black window, which then quickly got overwritten. So it looks like a brief black flash. I added code that prevents this black flash.

jacobvosmaer:
Thanks @prodhe 

While using this myself I encountered an (aesthetic) bug where there could be a drawRect call before devdraw had received any draw operations from its client (e.g. Acme). This would lead to a black window, which then quickly got overwritten. So it looks like a brief black flash. I added code that prevents this black flash.

jacobvosmaer:
Thanks @prodhe 

While using this myself I encountered an (aesthetic) bug where there could be a drawRect call before devdraw had received any draw operations from its client (e.g. Acme). This would lead to a black window, which then quickly got overwritten. So it looks like a brief black flash. I added code that prevents this black flash.

jxy:
I didn't want to compete with you, but I feel there is just too much useless code in the current cocoa-screen and I made another rewrite, see #194.

jacobvosmaer:
@jxy Neat! I'll try it out.

jacobvosmaer:
While testing #194 I found a small bug in this PR (191): after disconnecting an external monitor, the Acme window does not get re-drawn automatically. I have to do something like a window resize (corner drag) to get it to update. I think this broke because the old renderer had a refresh timer and my code doesn't. Should be able to fix that.

rsc:
From the discussion here and on #194 it looks all the focus is on #194, so I'm going to close this one. Thanks to both you and @jxy for fixing this!

PR #196, opened 2018-10-25
fontsrv: disable font smoothing on osx

jxy:
macOS Mojave version 10.14 starts to disable font smoothing.
We disable font smoothing for OSX_VERSION >= 101400 to match the
system default font rendering.
It also makes the font rendering on macOS similar to that on X11.
fontsrv


jacobvosmaer:
Thanks! This fixes #198.

PR #203, opened 2018-11-16
INSTALL: use cc to check for fontsrv on non-Darwin

jxy:
FreeBSD 11.2 by default does not have gcc.


rsc:
This is probably going to break some awful system like Solaris but we'll see...

PR #121, opened 2017-11-02
Support Mac IME

pocket7878:
Roughly implemented Mac IME support to devdraw:

- based on [tiancaiamao/devdraw](https://github.com/tiancaiamao/devdraw) 's Chinese IME support, and improve IME handling 
- Implement NSTextInputClient on devdraw's contentView.
- Change keyevent buffer to linked list from ring-buffer to avoid ring-buffer overflow.

Currently supported inline translation, but it needs patching to some application(sam).
~~So I'm thinking about displaying temporary input buffer on devdraw layer and send text to application after user submitted text.~~

- Acme works fine (with my local Japanese IME) without any modification
- Sam works with little modification to strict check Cut, Paste, Copy Kcmd

- 9term works with cooked mode.

PR #211, opened 2018-11-17
samterm: Strict check Kcmd non-typing character to handle Mac IME.

pocket7878:
Samterm should strict check non-typing character Kcmd to handling Mac IME marked range.

Without this patch illegal range deleted:
sam-ime

With this patch:
sam-ime-fix

This change was included in my closed patch of support Mac IME #121 
But not included changes of #194, so I sent this again 😀 

And this patch also solves can't input Zenkaku alphabet character(e.g `ｈ`)  in sam.

### DIFF
--- a/src/cmd/samterm/main.c
+++ b/src/cmd/samterm/main.c
@@ -512,10 +512,11 @@ nontypingkey(int c)
 	case PAGEUP:
 	case RIGHTARROW:
 	case SCROLLKEY:
+	case CUT:
+	case COPY:
+	case PASTE:    
 		return 1;
 	}
-	if(c >= Kcmd)
-		return 1;
 	return 0;
 }
 


jxy:
So sam was generating extra back space keys?  What was going on with out this patch?

pocket7878:
@jxy 
In Japanese IME, some typing are simply generate single character:
When I type `a` on my keyboard, it will generate `あ`.

But some typing are generate temporary character before sent final character:
When I type 'ha' on my keyboard, it will sent first `ｈ` (and it is not ascii `h`, but Zenkaku-h character') and send backspace and resend `は`.

Without this patch, sam will ignore `ｈ` cause it handles as a non-typing key. And will cause delete illegal range.
So, when I just typing  `あ`(`a`) after first string, this problem is not happen.

`ああああ` Gif:
sam-ime-aa
`は` Gif:
sam-ime-ha

PR #212, opened 2018-11-17
devdraw: Handle windowDidResize to support non-live window resize.

pocket7878:
Some window management tools (e.g [Moom](https://manytricks.com/moom/) which I'm using on my mac) using Accessibility API to support resizing window from keyboard.
And current Metal cocoa screen (Implemented in #194) only redraw after live resize event to improve performance.
But window resizing through accessibility API does not sent live-resize event.
So I'm add a handler for windowDidResize event, and check inLiveResize flag to skip inefficient redraw.

### DIFF
--- a/src/cmd/devdraw/cocoa-screen-metal.m
+++ b/src/cmd/devdraw/cocoa-screen-metal.m
@@ -384,6 +384,13 @@ struct Cursors {
 	return YES;
 }
 
+- (void)windowDidResize:(NSNotification *)notification
+{
+	if(![myContent inLiveResize] && img) {
+		resizeimg();
+	}
+}
+
 - (void)windowDidBecomeKey:(id)arg
 {
         [myContent sendmouse:0];
@@ -1148,7 +1155,6 @@ resizewindow(Rectangle r)
 
 		s = [myContent convertSizeFromBacking:NSMakeSize(Dx(r), Dy(r))];
 		[win setContentSize:s];
-		resizeimg();
 	});
 }
 


src/cmd/devdraw/cocoa-screen-metal.m
```diff
@@ -558,6 +565,15 @@ - (void)resetCursorRects {
 	[self addCursorRect:self.bounds cursor:currentCursor];
 }
 
+- (void)windowDidResize:(NSNotification *)notification
+{
+	if([self inLiveResize]) {
+		LOG(@"Skip resizeimg in live resize");
```
mattwidmann: Is this useful? Seems like it would overwhelm the log whenever a user-driven resize is active.

src/cmd/devdraw/cocoa-screen-metal.m
```diff
@@ -197,6 +197,7 @@ + (void)makewin:(NSValue *)v
 	[win setContentView:myContent];
 	[myContent setWantsLayer:YES];
 	[myContent setLayerContentsRedrawPolicy:NSViewLayerContentsRedrawOnSetNeedsDisplay];
+	[[NSNotificationCenter defaultCenter] addObserver:myContent selector:@selector(windowDidResize:) name:NSWindowDidResizeNotification object:nil];
```
mattwidmann: Should this pass `win` as the object? I wonder why `myContent` would work.

src/cmd/devdraw/cocoa-screen-metal.m
```diff
@@ -373,6 +374,12 @@ - (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)theAppl
 	return YES;
 }
 
+- (void)windowWillClose:(NSNotification *)notification
```
mattwidmann: Does this need a notification registered for it, too? Or is this enabled by default when asking for another notification on the window?

src/cmd/devdraw/cocoa-screen-metal.m
jxy: I don't think we need to register this event for the view.  Just using the NSApplicationDelegate method should be enough.

jxy:
This may overlap with other resizing event.  Perhaps we need to remove events that calls to resizeimg?

src/cmd/devdraw/cocoa-screen-metal.m
pocket7878: @mattwidmann Thank you for your review! I'll remove this debug Log.

src/cmd/devdraw/cocoa-screen-metal.m
pocket7878: Thank you for your review!
@jxy If I use NSWindowDelegate method on AppDelgate,
How can I check inLiveResize flag of DevDrawView?

pocket7878:
@jxy Thank you for your review
I've added some LOG line to other events `viewDidEndLiveResize`, `viewDidChangeBackingProperties`.

When I resize by mouse, `resizeimg` runs only once from `viewDidEndLiveResize`, cause I cheked `inLiveResize` flag in inside `windowDidResize` callback.
When I resize form keyboard `resizeimg` runs only once from `windowDidResize`, cause that event is not live resize event.
So I think there might be no overlap.

src/cmd/devdraw/cocoa-screen-metal.m
```diff
@@ -373,6 +373,13 @@ - (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)theAppl
 	return YES;
 }
 
+- (void)windowDidResize:(NSNotification *)notification
+{
+	if(![myContent inLiveResize]) {
```
jxy: I just realized that we probably need to check if img is nil before calling resizeimg.  It may get resized before attachscreen.

jxy:
I think setContentSize would trigger windowDidResize.  Please verify.  Probably we can just remove the resizeimg call after the call to setContentSize in function resizewindow.

jxy:
It's probably better to squash your commits.

pocket7878:
@jxy Thank you for your review! I just update codes & squashed

mkmik:
Compiling this on macos 10.14.5 I get:

```
9c -DOSX_VERSION=101405 -fobjc-arc -o cocoa-screen-metal-objc.o cocoa-screen-metal.m
cocoa-screen-metal.m:393:9: error: use of undeclared identifier 'windowDidBecomeKey'
cocoa-screen-metal.m:1274:2: error: expected '}'
cocoa-screen-metal.m:388:1: note: to match this '{'
cocoa-screen-metal.m:1274:2: error: missing '@end'
cocoa-screen-metal.m:139:1: note: implementation started here
mk: 9c -DOSX_VERSION=101405 -fobjc-arc ...  : exit status=exit(1)
```

jxy:
Yes.  There is an extra `{` after the `if` statement.

pocket7878:
@mkmik Thank you for your reporting! Fixed :)

rsc:
Committed, thanks.

Issue #208, opened 2018-11-16
devdraw: screen goes black after full-screen switch on macOS 10.14


rsc:
We don't know why but we merged #194 anyway since occasional black screens was better than nothing working at all.

/cc @jxy

jxy:
Let's gather some facts.  Do all of you having black screens have a discrete GPU?

mattwidmann:
I don't see a black screen when switching to full-screen mode. I'm running acme/devdraw on:

macOS 10.14.1 (18B75)
MacBook Pro (13-inch, 2017, Two Thunderbolt 3 ports)
Intel Iris Plus Graphics 640 1536 MB

pocket7878:
I have same problem on my MBP:

macOS: 10.14.1
plan9port 2607cc5 (master right now)
MacBook Pro (Retina, 15-inch, Mid 2014)
NVIDIA GeForce GT 750M 2048 MB
Intel Iris Pro 1536 MB

larryr:
not sure if this is related as it is not only on full-screen.
I run /bin/plot which draws a plot on the screen successfully.  If you expand the windows size just a little toward left (e.g. with mouse grab left side of window and slight move to left expanding the size. Then do the same on the right side just a small amount to the right.  When you release the mouse button the window will go black.

Sometimes if you only resize each direction just a few pixels (or so) you'll get an all white screen although you see a black band around part.  This is pretty consistent. and seems limited to resizing one direction first then the other.

up/down does the same.

All this on a secondary screen. Setup in non-mirror mode (e.g. laptop and external monitor active and separate screen content.

Condition happens on either display -- external or laptop.  Condition happens on laptop screen when external monitor is disconnected, as well.

This on: 
MacBook Air 6,1 (older, intel Core i7 1.7Ghz 2-core) 
Intel HD Graphics 5000, 1536MB VRAM,  devid:0x0a26, RevID:0x0009
Metal: Supported, feature set maxOS GPUFamily1 v4
MacOS: 10.14 (18A391)
Darwin: 18.0.0
plan9port: 2607cc5

hope this helps.  Mojave changes have been very welcomed.

larryr:
It’s actually working pretty well other than the redraw problem I mentioned. 

....Larry

jxy:
Yes.  It only works with the new devdraw for mojave, but not with the previous cocoa devdraw.

lufia:
I have same issue.

* macOS: 10.14.2
* plan9port 2607cc5
[...]

PR #213, opened 2018-11-17
devdraw: call resizeimg on NSWindowDidEnterFullScreenNotification callback.

pocket7878:
Calling resizeimg on NSWindowDidEnterFullScreenNotification callback to redraw.
It solves black fullscreen problem on my laptop.
#208


pocket7878:
After sent this PR, I've found Ive still got a same problem after more repeat testing.
I'm sorry to have jumped to the conclusion..

PR #216, opened 2018-11-20
acme: macOS (emacs?) keyboard shortcuts

ruda:
This is a work in progress  to implement more keyboard shortcuts to ACME, emulating the behavior of macOS (should I say emacs?) when editing on the terminal. For example:

Command+A will select all text, Control+A will put the cursor on the beginning of line, Control+D will delete one character forward, etc. This will work on Windows too with the Windows Key and I think it will do the same on Linux.

Let me know if there is any interest in doing this kind of modification in ACME. I know the implementation is bugged and I don't have much experience with ACME code, so I need some help here.


kare:
Acme is intentionally mouse driven. For more information see http://9p.io/wiki/plan9/Mouse_vs._keyboard/index.html

jxy:
And the trackpad on the MacBooks works well too, especially when it's literally under your palms.

PR #217, opened 2018-11-21
libdraw,devdraw: fix compatibility with old 16x16 cursor protocol

fhs:
Some libraries that depend on devdraw don't know about
32x32 cursor -- mainly 9fans.net/go/draw.


jxy:
I don't know if there is any other code than the go/draw that depends on this.  Perhaps it is just easier to update go/draw than adding a compatibility layer?

fhs:
I think 32x32 cursor was originally intended to be supported by a new message type (Tcursor2 was added to drawfcall.h in 8581c2b56763d7787604c8c833d2bd78bdc6a466). It somehow was forgotten in the actual implementation.

Issue #219, opened 2018-11-27
acme: crash when using Edit X/foo/ ,|bar


jacobvosmaer:
To reproduce:

```
B /tmp/a
B /tmp/b
```

Then in any tag, `Edit X/tmp/ ,|fmt`. I expect this to pipe each file through `fmt` but instead Acme just crashes. The crash doesn't happen if the `X` query matches just one window.

jacobvosmaer:
This might be a duplicate of #9

jxy:
I think plan 9 have the same problem.  Apparently `winunlock` is called without a prior `winlock` at `ecmd.c:/^runpipe/+/winunlock` with the X command.

jxy:
Please test #222

jacobvosmaer:
Cross-posting from #222: it works!

PR #222, opened 2018-11-28
acme: fix X, Y command crashing with extra locks

### LINKED ISSUE #9: cmd/acme: Edit X/.../ broken
Acme's implementation of the X sam looping construct crashes with error:

  qunlock pc=0x1079c6b30 owner=0 self=108124000 oops

How to reproduce?
1. Execute 'New foo'.
2. Execute 'Edit X/foo/ >wc'.

jxy:
When mouse issues a command, the window that has the mouse is locked
with winlock, then the command is run, and then winunlock.  Runpipe
is coping this mechanism by sequentially calling incref, run (which
contains a winclose that does a decref, winunlock, and finally
winlock again.  This is fine for the usual pipe_cmd, but for X_cmd,
all except one window does not have the lock to begin with.  Acme
would crash when such X, Y command were run and winunlock were
called on a window without a prior winlock.

This patch fix the issue by adding extra winunlock/winlock in X_cmd
to deal with existing locks, and applying winlock/winunlock for each
matching window in filelooper.

Fixes #9 #219

### DIFF
--- a/src/cmd/acme/dat.h
+++ b/src/cmd/acme/dat.h
@@ -508,7 +508,6 @@ enum	/* editing */
 	Collecting
 };
 
-uint		globalincref;
 uint		seq;
 uint		maxtab;	/* size of a tab, in units of the '0' character */
 
--- a/src/cmd/acme/ecmd.c
+++ b/src/cmd/acme/ecmd.c
@@ -582,9 +582,11 @@ x_cmd(Text *t, Cmd *cp)
 int
 X_cmd(Text *t, Cmd *cp)
 {
-	USED(t);
-
+	if(t && t->w)
+		winunlock(t->w);
 	filelooper(cp, cp->cmdc=='X');
+	if(t && t->w)
+		winlock(t->w, cp->cmdc);
 	return TRUE;
 }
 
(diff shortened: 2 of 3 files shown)


jacobvosmaer:
I am the one who opened #219. This PR fixes the problem! Thanks @jxy

jacobvosmaer:
Bump!

I have been using a custom branch of Acme with this fix in it for almost a year now.

rsc:
I understand the rationale for doing the unlock+relock dance but I think the ref counts are still needed. The point is to account for the window being in the loopstruct list and make sure it does not get freed while in the list. For example if you have two windows and run Edit |sleep 5 and then during the first window's sleep 5, the second window is closed (by a 9P operation), it is important not to free that window until the X finishes. Otherwise the X on the second window will be using a dangling pointer. No?

jxy:
That's right.  I missed that.  Thanks.  I'll look at it later.  Perhaps adding the ref counts back with more checking.

jacobvosmaer:
Maybe something like https://github.com/9fans/plan9port/pull/330/files ?

jacobvosmaer:
Thanks @rsc!

PR #226, opened 2018-12-17
Map stack explicitly in openbsd to avoid segfaults (credit: Ori_B)

neutralinsomniac:


goleo108:
err, shouldn't common code go to src/libthread/OpenBSD.c? also why not reuse MAP_STACK (non-standard extension) on Linux, FreeBSD and DragonFly BSD?

PR #228, opened 2018-12-29
Update TODO

ghost:
added request to generate an external library dependency list


rsc:
Removed TODO instead.

PR #231, opened 2019-01-07
prefer integrated GPU for macOS metal rendering

schultzor:
hi, apologies in advance for how rough this might be: I noticed that acme was marked "Yes" in the "Requires High Perf GPU" column in macOS' Activity Monitor app on my dual-GPU macbook. Selecting a low-power integrated GPU seems to change this behavior, I'm unaware if there's a reason to use the default/high-performance GPU for devdraw.


jxy:
There is no reason to use a high perf GPU.  I just don't have a discrete GPU machine to test it.

While you are at it, would this work for displays driven by external GPUs?  I guess we can wait for someone who has an external GPU to submit more PRs.

schultzor:
That's a good question. I learned just enough about Metal to make this PR, and don't have an e-GPU handy to test that case.

rsc:
I'm willing to try this and see.

Issue #232, opened 2019-01-12
Acme macos cursor not focussed after task switch, sometimes


jacobvosmaer:
I use acme on macos 10.14.1, the new metal backend. Sometimes when I task switch from another app to Acme, the cursor is somehow not focussed. I type on the keyboard but not text appears. Only when I move the mouse does Acme focus on the right window.

I'm not sure how to reproduce this (yet) but it happens quite regularly.

jacobvosmaer:
I think I may have found a way to reproduce.

There are areas in an Acme window where keypresses get ignored. The border between two "typable" areas for instance. I can reproduce if I do this:

- move mouse in Acme to area where keypresses get ignored
- switch to other app
- within other app, move mouse to point where acme _should_ respond to the keyboard
- switch back to acme
- press keys

Result: nothing happens until I move the mouse. I wonder if we're missing some sort of mouse event after switching back.

jxy:
Yes, I understand that the mouse location is not updated (the new mouse location is not communicated to acme) after switching back to acme.  I will take a look at it later.

PR #239, opened 2019-02-20
devdraw: respond to windowDidBecomeKey on darwin

jacobvosmaer:
Fixes bug where devdraw does not "notice" mouse position after task
switch. Fixes https://github.com/9fans/plan9port/issues/232.

In Acme, which uses "focus follows mouse", this bug is particularly annoying. The effect is that I often have to wiggle my mouse (or touch the trackpad) after switching back to Acme with Cmd-Tab.

### DIFF
--- a/src/cmd/devdraw/cocoa-screen-metal.m
+++ b/src/cmd/devdraw/cocoa-screen-metal.m
@@ -373,6 +373,11 @@ struct Cursors {
 	return YES;
 }
 
+- (void)windowDidBecomeKey:(id)arg
+{
+        [myContent sendmouse:0];
+}
+
 @end
 
 @implementation DevDrawView


jacobvosmaer:
What do you think about this fix @jxy?

src/cmd/devdraw/cocoa-screen-metal.m
```diff
@@ -373,6 +373,11 @@ - (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)theAppl
 	return YES;
 }
 
+- (void)windowDidBecomeKey:(id)arg
+{
+        [myContent sendmouse:0];
```
jacobvosmaer: I'm not sure what sending `0` means. The old devdraw called a function `sendmouse()` without arguments.

But, this seems to work, and I see `[myContent sendmouse:0]` in other places as well.

jxy: 0 means mouse buttons are all released, so it will update the location and tell the application no mouse buttons are pressed.

jxy:
Yes.  This is all we need.

jacobvosmaer:
I've updated the PR description with a bit more context about the bug.

PR #242, opened 2019-03-01
Various Tweaks and Improvements to Sam

ryukoposting:
I've copied over my modifications of Sam/Samterm from [my fork of plan9port on GitLab](https://gitlab.com/ryukoposting/plan9port). The changes I've made are:

 - Added a new argument to samterm (and to Sam, but Sam merely passes the argument up into Samterm), `-i[nspaces]`. Giving `-i` with a number will cause tab keystrokes to insert the specified number of spaces instead. For example, `-i4` will make pressing the 'tab' key result in the insertion of 4 spaces, instead of a tab character. This makes life easier when working on projects where tabs are not the standard indenting character.
- The behavior of the delete key has been altered in samterm. Instead of behaving like the backspace key (deleting the character before dot), the delete key now will delete the character immediately following dot.
- The home key now moves dot to the beginning of the current line, instead of the beginning of the current file. The home key is also now indentation sensitive- if there are non-indenting (that is, not `'\t'` or `' '`) characters before dot, the home key will move dot to be directly before the first non-indenting character of the line. If all characters on the current line before dot are either `'\t'` or `' '`, then the home key will move dot to the beginning of the line.
- The end key moves dot to the end of the current line, instead of the end of the current file.

Obviously, my modifications may introduce behaviors that don't match the original Sam. I'm not sure how much you guys are interested in keeping the plan9port programs 100% faithful to the originals, but I think these changes make Sam much more practical for everyday use, without fundamentally altering its design.

### DIFF
--- a/src/cmd/sam/sam.c
+++ b/src/cmd/sam/sam.c
@@ -43,7 +43,9 @@ main(int _argc, char **_argv)
 	volatile int i, argc;
 	char **volatile argv;
 	String *t;
-	char *termargs[10], **ap;
+	char *termargs[32], **ap;
+	char *p;
+	int ilen;
 	
 	argc = _argc;
 	argv = _argv;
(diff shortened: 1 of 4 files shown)


rsc:
We're not going to diverge from original sam, sorry.

PR #248, opened 2019-03-19
port of Plan 9 winwatch(1)

markvanatten:
These are a port of Plan 9 winwatch(1) and its man page. The work was mostly done by Fazlul Shariar:
[https://github.com/fhs/misc/blob/master/cmd/winwatch/winwatch.c](url)
I modified it to repair memory leaks and to deal with certain X errors arising from its asynchronous nature. I added two options: with -n, the label is defined by the window's name instead of its class, and with -s the labels are sorted by alphabet (case insensitive, unicode aware) instead of by order of appearance.


src/cmd/winwatch.c
```diff
@@ -0,0 +1,543 @@
...
+
+void usage(void)
+{
+    fprint(2,
+           "usage: winwatch [-e exclude] [-W winsize] [-f font] [-n] [-s]\n");
+    exits("usage");
+}
+
+
+void main(int argc, char **argv)
```
fhs: Use Plan 9 C style, where the function name always start on a new line. It makes it easier to find function definitions. Here, `g ^main` would tell me where all the main functions are in this directory.

src/cmd/winwatch.c
```diff
@@ -0,0 +1,543 @@
...
+    }
+    return 0;
+}
+
+int winlabelcmp(const void *w1, const void *w2)
+{
+    const Win *p1 = (Win *) w1;
+    const Win *p2 = (Win *) w2;
+    int rc;
+    if ((rc = _cistrcmp(p1->label, p2->label)) < 0)
```
fhs: Just return the return value of `_cistrcmp`. You don't need any `if` statement.

markvanatten:
Thanks. You're right, I think, that it should be in $PLAN9/src/cmd/rio. I'll follow up the other suggestions also, but will wait a few days to see if any others come in.

Issue #247, opened 2019-03-18
9l doesn't support FreeBSD 12


francisconti:
The 'case $tag' in 9l for FreeBSD needs to be updated to deal with FreeBSD 12. Changing "1[0-1]" to "1[0-2]" is the obvious trivial way to do this, but should we really need to revisit this every major version? It seems like it'd be better to change that entire sub-case to the default (I'm not sure what the behavior of FreeBSD < 5.2 is here, but given that would be before 2004-01, I'm not sure we ought to worry too much).

Happy to make submit either change, but would appreciate guidance from the maintainers.

jxy:
Actually do we need `-lpthread`?  It looks like `libthr` is automatically linked for me and I have never had a problem with `9l`.

francisconti:
With stock p9p and a freshly-installed FreeBSD 12, INSTALL fails when 9l can't find some pthread symbol (pthread_create, IIRC). Changing "1[0-1]" to "1[0-2]" let installation complete.

I'm curious how you're getting libthr linked. How are you determining that?

jxy:
I just realized that I did make the change to `9l` and I completely forgot about it.  I guess change the pattern to be `1[0-9]*` to catch all version 10 or later would be good.

jxy:
Looking through the diffs, I also made some changes to `libmach/FreeBSD.c`
```
diff --git a/src/libmach/FreeBSD.c b/src/libmach/FreeBSD.c
index 7cf5c1f4..d6b00de5 100644
--- a/src/libmach/FreeBSD.c
+++ b/src/libmach/FreeBSD.c
@@ -30,8 +30,8 @@ struct PtraceRegs
        int pid;
 };
 
-static int ptracerw(Map*, Seg*, ulong, void*, uint, int);
-static int ptraceregrw(Regs*, char*, ulong*, int);
+static int ptracerw(Map*, Seg*, u64int, void*, uint, int);
+static int ptraceregrw(Regs*, char*, u64int*, int);
 
 void
 unmapproc(Map *map)
@@ -92,7 +92,7 @@ detachproc(int pid)
 }
 
 static int
-ptracerw(Map *map, Seg *seg, ulong addr, void *v, uint n, int isr)
+ptracerw(Map *map, Seg *seg, u64int addr, void *v, uint n, int isr)
 {
        int i;
        u32int u;
@@ -155,7 +155,7 @@ static char *freebsdregs[] = {
        "GS",
 };
 
-static ulong
+static u64int
[...]

PR #249, opened 2019-03-20
9l: support FreeBSD ≥ 12 (fixes issue #247)

francisconti:
Per the discussion on issue #247: makes 9l correct for FreeBSD 12, and we won't have to revisit it until FreeBSD 20, modulo other changes.

### DIFF
--- a/bin/9l
+++ b/bin/9l
@@ -19,7 +19,7 @@ case "$tag" in
 	5.2.*)
 		extralibs="$extralibs -lkse"
 		;;
-	[5-9].*|1[0-1].*)
+	[5-9].*|1[0-9].*)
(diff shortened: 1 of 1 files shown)


jxy:
You could just make the pattern `[0-9][0-9]*`.

wcheswick:
This hurts.  I am looking forward to the fix.

ches

0intro:
@dancrossnyc You changed the PR title, but not the commit title. What I do in this case is to apply the change manually, amend the commit to edit the message, then push.

dancrossnyc:
Ugh. Very well; thanks for the heads up, David. It sounds laborious but
doable.

PR #251, opened 2019-04-01
acme: Update tag after receiving menu/nomenu control event

mkhl:
After acme receives a `menu` or `nomenu` control event, it doesn’t update the tag.
Since these events potentially modify the tag (Undo, Redo and Put are only automatically included if `menu` is active), it should.


rsc:
I'm not 100% sure about this but I'm willing to try it and see if anything breaks.

PR #252, opened 2019-04-11
samterm: add mousewheel scrolling support

ghost:
Hi,

To clarify the intent, line 156 and 164 in scroll.c were added to limit the change to the scroll() function as the 'but' argument is passed to the scrorigin() function and I did not see any good reason to modify it.

Thanks
Philippe


dancrossnyc:
I know there is some sensitivity about making changes to sam, so I added the relevant reviewers.

rsc:
When I tested the original mousewheel support long ago it hung sam under conditions that I never worked out. So I have been reluctant to add it, for a long time.

PR #253, opened 2019-04-12
devdraw/x11: fix modifier key handling for some XkbOptions

jxy:
Certain XkbOptions in X11 would change keysyms for modifier keys
between the key press and key release.

For example, under the XkbOptions "grp:shifts_toggle", though shift
keys remain Shift_L/R when pressed, they become ISO_Group_Next/Prev
when released.

This behavior makes devdraw unable to detect the release event
correctly and as a result mouse button 1 click always interpreted
as button 3 event after a shift key is used.

### DIFF
--- a/src/cmd/devdraw/x11-screen.c
+++ b/src/cmd/devdraw/x11-screen.c
@@ -291,6 +291,8 @@ xloop(void)
 	}
 }
 
+static int kcodecontrol, kcodealt, kcodeshift;
+
 /*
  * Handle an incoming X event.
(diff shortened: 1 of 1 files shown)


rsc:
I just checked in a whole bunch of changes to the x11 code.
If you'd like to rebase this PR on top of them and check that it still works, I'd be happy to merge it.
Thanks.

(Any PRs fixing bugs in what I just checked in, whether for macOS or X11, are welcome too!)

PR #254, opened 2019-04-12
fontsrv/x11: allow hinting and disable autohint only

jxy:
Some truetype fonts have good manual hinting.
Ignoring hinting makes the font render badly on low resolution screens.
This commit only disables the freetype autohinter, and allows hinting.


aksr:
I agree with you.

PR #258, opened 2019-05-02
acme: send/receive incl attr in plumb message

scgehin:
Acme will plumb files instead of opening when possible. This makes it so that new windows opened do not inherit Incl directories while the plumber is running. Fixed by adding an "incl" attribute to the plumb message that contains the list of directories, delimited by '|'.

### DIFF
--- a/src/cmd/acme/look.c
+++ b/src/cmd/acme/look.c
@@ -78,6 +78,36 @@ startplumbing(void)
 	threadcreate(plumbthread, nil, STACK);
 }
 
+void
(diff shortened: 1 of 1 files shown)


robpike:
Not sure I want to mix worlds like this, or to put it another way I'm not convinced this is the right approach.

dancrossnyc:
In light of Rob's comment, I'm going to close this. If another solution is in the offering at some point that doesn't change the plumb protocol, I'm sure we'd be happy to have a look at it.

Issue #260, opened 2019-05-04
cmd/devdraw: color space issues in full screen with macos "Night Shift"


jacobvosmaer:
macos 10.14.3, plan9port 9179fdaa

I use the "Night Shift" feature available in recent macos versions, which (drastically) changes the color temperature of the display after sunset to go easier on the eyes. It gives everything a warm yellowish tint.

This mostly works as expected with devdraw (acme) except when I use full screen mode. In that case I intermittently but quite often see a jarring effect where the white outline of the mouse cursor is at normal daytime color temperature. It's probably the entire cursor that is wrong but it's most visible on the white outline which becomes bright / blueish in comparison to the rest of the display.

I have not been able to make a screenshot because pulling up the screenshot app seems to reset some display state, and I'm not sure if the screenshots reflect the night shift color temperature to begin with. (I _think_ the screenshots turn out neutral.)

Edit: I think this is not just about the cursor, it's about the color space of the whole window (screen), it's just most visible around the cursor.

jacobvosmaer:
I need to do more work to try and reproduce this. It's not enough to manually turn on night shift.

jacobvosmaer:
It looks like you also need to enable the "automatically adjust brightness" feature, and trigger brightness adjustments (e.g. by blocking light from reaching the sensor used for that).

jacobvosmaer:
OK so not related to auto brightness after all.

- enable night shift (make sure color temperature is on the warm side)
- start acme (as normal window)
- make sure there is a acme window you can type text in
- switch acme to full screen
- type more text

Result: cursor color space gets messed up.

I tried disabling the custom cursor by commenting out `- (void)resetCursorRects`. With that function removed, I get the normal cursor again, but the color problem still happens.

jacobvosmaer:
I also see a jump in the window color after switching to full screen. Maybe the cursor is a red herring? It might be that something is wrong with the colors overall, and it's just most noticeable on the cursor.

jacobvosmaer:
What do you think about this change @jxy?

One way to reproduce this bug is to turn on night shift ("manually enable until sunrise"), switch acme to full screen and then typing and moving the mouse.

jxy:
@jacobvosmaer do you experience the same thing when #215 is applied?

jacobvosmaer:
@jxy it looks like #215 also fixes this issue!

jxy:
[...]

PR #215, opened 2018-11-19
devdraw: cocoa metal screen uses a dirty hack to make everything smooth

jxy:
You are going to hate this.

However, this one line fixes #207 #208 #214.

PR #261, opened 2019-05-05
cmd/devdraw: set color space on CAMetalLayer

jacobvosmaer:
Fixes https://github.com/9fans/plan9port/issues/260.

The macos "Night Shift" feature (automatic color temperature changes to try and match daylight / artificial light conditions) works as expected with windowed devdraw windows. But the moment I go to full screen (Cmd-F) and type something, the color space gets messed up.

This change, which I discovered through trial and error, seems to fix the problem.

### DIFF
--- a/src/cmd/devdraw/cocoa-screen-metal.m
+++ b/src/cmd/devdraw/cocoa-screen-metal.m
@@ -217,6 +217,7 @@ threadmain(int argc, char **argv)
 	layer.pixelFormat = MTLPixelFormatBGRA8Unorm;
 	layer.framebufferOnly = YES;
 	layer.opaque = YES;
+	layer.colorspace = CGColorSpaceCreateWithName((__bridge CFStringRef)NSDeviceWhiteColorSpace);
 
 	renderPass = [MTLRenderPassDescriptor renderPassDescriptor];
 	renderPass.colorAttachments[0].loadAction = MTLLoadActionDontCare;


src/cmd/devdraw/cocoa-screen-metal.m
```diff
@@ -217,6 +217,7 @@ + (void)makewin:(NSValue *)v
 	layer.pixelFormat = MTLPixelFormatBGRA8Unorm;
 	layer.framebufferOnly = YES;
 	layer.opaque = YES;
+	layer.colorspace = CGColorSpaceCreateWithName((__bridge CFStringRef)NSDeviceWhiteColorSpace);
```
jacobvosmaer: 1. I don't know what I'm doing with object ownership / memory management here, but I think NSDeviceWhiteColorSpace is a global so we don't have to become owner of it
2. It seems awkward to have to use a "NS" constant as an input to a "CG" function. Maybe there is a better way?

jacobvosmaer: 3. I also can't explain why what I am doing here is "right", it just seems to work. This uses the same color space we use when instantiating the cursor images.

jacobvosmaer:
It looks like #215 also fixes #260. It sounds like #215 fixes more problem than just the color space, so I am going to close this.

PR #264, opened 2019-05-18
9l: Update and near-future-proof Linux version check

escondida:
The Linux kernel had another major version bump to 5, meaning that 9l wasn't picking up -lpthread. This patch should cover future versions, unless the entire version scheme or the name of the threading library changes.


rsc:
Fixed by 13d40aa64.

PR #284, opened 2019-08-24
Fix memory leak when dial opens a regular file

jvd23:


nsajko:
@jvd23  Take a look at one of the following links. The first line of git commit messages for plan9port has a conventional form. (The same form as for Go.)

https://github.com/golang/go/wiki/CommitMessage

https://golang.org/doc/contribute.html#commit_messages

PR #285, opened 2019-09-01
acme: show Get command not only for dirs' windows

rslnx:
Handy to have Get command always available for files, especially after `go fmt`.


rsc:
Sorry, but no. Plan9port is meant to be a port of the tools, not a place to revisit design decisions. 
If you want Get in a file tag, type "Get". (That's the great thing about acme!)

As far as go fmt is concerned, you might be interested in 9fans.net/go/acme/acmego.

PR #287, opened 2019-09-11
hoc: don't nest calls to follow() when lexing ++/+= and --/-=

deepcube:
The code had a nested use of the follow() function that could cause +=+
and -=- to register as ++ and --.  The first follow() to execute could
consume a character and match and then the second follow() could consume
another character and match.  For example i-=-10 would result in a syntax
error and i-=- would decrement i.

### DIFF
--- a/src/cmd/hoc/hoc.y
+++ b/src/cmd/hoc/hoc.y
@@ -215,8 +215,8 @@ yylex(void)		/* hoc6 */
 		return STRING;
 	}
 	switch (c) {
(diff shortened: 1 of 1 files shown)


nsajko:
You should probably change the capital "D" to lowercase "d" in the commit message and pull request title, for consistency with existing commits.

deepcube:
> You should probably change the capital "D" to lowercase "d" in the commit message and pull request title, for consistency with existing commits.

Done^wdone

PR #288, opened 2019-09-18
plumber: add -f (foreground) option

eraserhd:
In MacOS, services run by launchd must run in the foreground, since
launchd manages forking and other resources.

### DIFF
--- a/man/man4/plumber.4
+++ b/man/man4/plumber.4
@@ -4,6 +4,9 @@ plumber \- file system for interprocess messaging
 .SH SYNOPSIS
 .B plumber
 [
+.B -f
+]
+[
 .B -p
 .I plumbing
 ]
@@ -23,7 +26,7 @@ in the format of
 Its services are posted via
 .IR 9pserve (4)
 as
-.BR plumb .
+.BR plumb ,
 and consist of two
 pre-defined files,
 .B plumb/send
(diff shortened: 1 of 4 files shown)


man/man4/plumber.4
```diff
@@ -23,8 +26,8 @@ in the format of
 Its services are posted via
 .IR 9pserve (4)
 as
-.BR plumb .
-and consist of two
+.BR plumb ,
+and consists of two
```
iburago: The subject is plural, so the predicate should remain to be “consist”, shouldn’t it?

src/cmd/plumb/fsys.c
```diff
@@ -199,7 +199,10 @@ startfsys(void)
 	if(post9pservice(p[1], "plumb", nil) < 0)
 		sysfatal("post9pservice plumb: %r");
 	close(p[1]);
-	proccreate(fsysproc, nil, Stack);
+	if (foreground)
```
iburago: Generally, no spaces are put before the opening parenthesis in the rest of the codebase.

man/man4/plumber.4
eraserhd: You are right.

eraserhd:
Amended the commit with above adjustments and changed the title to match requirements from the list.

PR #303, opened 2019-12-22
libthread: use movw with 16-bit segment registers on OpenBSD/i386

goleo108:


nsajko:
@goleo108 Take a look at one of the following links. The first line of git commit messages for plan9port has a conventional form. (The same form as for Go.)

https://github.com/golang/go/wiki/CommitMessage

https://golang.org/doc/contribute.html#commit_messages

goleo108:
@nsajko note that I am completely ignorant on assembler topics, so I can't describe the actual problem, the patch is coming from @sthen

dancrossnyc:
Hmm. I would submit that the FreeBSD code is wrong, and should be fixed and this change reverted.

PR #304, opened 2019-12-25
Two commits that fix quite a few of libhtml memory leaks

nsajko:


src/libhtml/build.c
```diff
@@ -2127,13 +2127,17 @@ getitems(ItemSource* is, uchar* data, int datalen)
 	outerps = lastps(ps);
 	ans = outerps->items->next;
 	/* note: ans may be nil and di->kids not nil, if there's a frameset! */
+	if (outerps->items != nil) {
```
dancrossnyc: Just free without the if please.

src/libhtml/lex.c
```diff
@@ -584,14 +587,22 @@ getplaindata(TokenSource* ts, Token* a, int* pai)
 		if(c != 0){
 			buf[j++] = c;
 			if(j == sizeof(buf)-1){
+				tmpS = s;
 				s = buftostr(s, buf, j);
+				if (tmpS != nil) {
```
dancrossnyc: Just `free(tmpS);` here.

src/libhtml/lex.c
```diff
@@ -584,14 +587,22 @@ getplaindata(TokenSource* ts, Token* a, int* pai)
...
+				}
 				j = 0;
 			}
 		}
 		if(c == '\n')
 			break;
 	}
+	tmpS = s;
 	s = buftostr(s, buf, j);
+	if (tmpS != nil) {
```
dancrossnyc: Same here. Just the free, no if.

dancrossnyc:
I made a few line comments, but the general point stands: instead of `if(p) { free(p); }` just `free(p);` throughout.

nsajko:
@dancrossnyc done.

src/libhtml/lex.c
```diff
@@ -734,15 +745,19 @@ getscriptdata(TokenSource* ts, int firstc, int starti, Token* a, int* pai)
 		if(c != 0){
 			buf[j++] = c;
 			if(j == BIGBUFSIZE-1){
+				tmpS = s;
```
dancrossnyc: This pattern is repeated enough that something like,

```
static Rune*
Buftostr(Rune* s, Rune *buf, int j)
{
        Rune *tmp = s;
        s = buftostr(tmp, buf, j);
        free(tmp);
        return s;
}
```

Might be warranted.

nsajko: Yeah, I changed the buftostr API, and now the change is much nicer. (Every call was a memleak.)

nsajko:
@dancrossnyc I think this is OK now?

dancrossnyc:
Indeed it is.

PR #305, opened 2019-12-25
Yacc fixes

nsajko:


src/cmd/yacc.c
```diff
@@ -1177,19 +1177,23 @@ setup(int argc, char *argv[])
...
 	vflag = 0;
 	dflag = 0;
+	fdflag = 0 != 0;
 	stem = 0;
 	stemc = "y";
 	foutput = 0;
 	fdefine = 0;
 	fdebug = 0;
 	ARGBEGIN{
+	case 'f':
```
dancrossnyc: Intuitively, I would think that `-f` would take a filename, not a file descriptor. Consider renaming this argument: perhaps `-F` would be sufficiently distinct?

src/cmd/yacc.c
```diff
@@ -1177,19 +1177,23 @@ setup(int argc, char *argv[])
...
+		fdflag = 0 == 0;
```
dancrossnyc: Surely just, `fdflag = 0;` ?

src/cmd/yacc.c
dancrossnyc: Though on further thought, file descriptors on plan9 appear in a filesystem namespace; under plan9, one wouldn't bother with this but would rather use some file name for an arbitrary FD: `/fd/12` for example.

On at least Linux, FreeBSD, OpenBSD and macOS, `/dev/fd` is contains files that seems to perform roughly the same function as on plan9. Given that those exist, one wonders whether this particular change is necessary; in the minimalist spirit of plan9 and its ports, I'd suggest that it is not.

src/cmd/yacc.c
dancrossnyc: (er, 1. But still, the boolean is a little too cute.)

nsajko: @dancrossnyc I think it would be best to do something like `#define True 1` (analogously for the other boolean value). Can you tell me if it should be `TRUE`, `True`, `true`, or just `1` directly?

nsajko: (I will remove this particular commit in a future force-push, I am asking for other commits.)

nsajko: For what it is worth, `True` and `TRUE` are used in that way in other places in p9p; and C++ and Go use `true`.

nsajko: <stdbool.h> also `define`s the lower case versions.

src/cmd/yacc.c
nsajko: @dancrossnyc I rm'd that commit now.

src/cmd/yacc.c
nsajko: I went with False and True enumeration constants.

nsajko:
It have a hunch that a lot of stuff in yacc.c could be improved with more usage of malloc. For example the complications around the temporary files would be extraneous if no temporary files were used. Also search for "someday, alloc should be used to allocate all this stuff".

rsc:
> It have a hunch that a lot of stuff in yacc.c could be improved with more usage of malloc. 

To be very clear, "improved" is _**not the goal**_.

The goal is to run code that was written a long time ago, with as few modifications (and new bugs introduced) as possible. If you find outright bugs, or if you find places where too-clever modern compilers will miscompile the code as written, we want to fix those. We do _not_ want to just clean up any of this code for the sake of cleaning it up. That's a non-goal. Thanks.

rsc:
I think this PR should probably be abandoned. It makes far too many unnecessary changes. If there are bugs to fix, fine. But for example there is no point in cleaning up "memory leaks" due to missing Bterms in a standalone command-line tool that runs quickly and exits.

nsajko:
@rsc I will make this clear in particular commit messages, but the reason for fixing memory leaks, if nothing else, is to enable in-process fuzzing (ref. Clang's libfuzzer). Fuzzers discovered the edge-case bugs like in e2b4c541274c25a1d3fa22e590d37ead8603520a or 7c0e88571902ede23b72dc3c9c8963bd1e1ee24f .

rsc:
This is too invasive. It is not a goal to refactor every command-line program to be able to be used with in-process fuzzing. Sorry. If you want to fuzz it, then it's fine to make a copy and hack it up, and it's also find to use out-of-process fuzzing like AFL. 

We're just not going to start modifying every single command-line tool to do this. It's too invasive and too prone to new bugs. Sorry.

I still think this PR sequence should be closed.

(Also note that when you do add new functions, the local style is to put the opening brace on a line by itself.)

nsajko:
@rsc Three commits here are strictly bug fixes (am I wrong?) and the one for trying to write tmp-files to /tmp instead of to the current directory could also be considered a fourth bugfix if yacc is thought of as a possible standalone tool instead of just being used to build plan9port. (Although perhaps a better approach would be to give the user the responsibility of choosing the temp dir?)

Do you want me to remove the non bugfix commits, so you could take a better look at these 3-4?

src/cmd/yacc.c
```diff
@@ -1936,6 +1937,11 @@ cstash(char *s)
 	return temp;
 }
 
+int
+isvalidchar(int i) {
+	return (i & ~0x7f) == 0;
```
nsajko: Hmm, I don't know why I made this 0x7f instead of 0xff, whic hseems like it would make more sense.

rsc:
If you can separate out the bug-fix commits, that would be great. Thanks.

nsajko:
@rsc done.

rsc:
Thanks. These fixes are great.

PR #306, opened 2019-12-25
Plug two libmach memory leaks

nsajko:


src/libmach/elf.c
```diff
@@ -287,6 +289,9 @@ elfinit(int fd)
 	return e;
 
 err:
+	if (p != nil) {
```
dancrossnyc: free(nil) is a no-op; why not just `free(p)` without the conditional?

nsajko: @dancrossnyc done as suggested

PR #309, opened 2019-12-28
cmd/htmlroff: fix buffer overflow in t2.c getqarg

nsajko:
This is actually from 2016:
https://plan9port-review.googlesource.com/c/plan9/+/1590

Change-Id: I6f2a3d71a9dd589eff7ab15b3c1d3997254b3c35


0intro:
Applied. Thanks.

PR #310, opened 2019-12-28
cb: fix a null pointer dereference

nsajko:
Just added a pair of parentheses. I also ran cb on cb.c to beautify the
code.

This is actually on Gerrit from 2016:
https://plan9port-review.googlesource.com/c/plan9/+/1574

Change-Id: I5e234adba0f95c13d6eecb121bf11bba4bf54566


0intro:
Applied. Thanks.

PR #311, opened 2019-12-28
ls: check that pointer is not nil before calling qsort

nsajko:
Passing a null pointer to qsort is an error in C (GCC and Clang agree
with the standards there, so this is no joke).

Change-Id: Ia2b015793a75ea4e85ae8f47da6beead9c4290e6


0intro:
Applied. Thanks.

PR #312, opened 2019-12-28
acme: do not pass null pointers where disallowed

nsajko:
The C standards disallow passing null pointers to memmove and memcmp.

Change-Id: I1c88c2adbc32a23ef742f206038b8f7c4e0540c7


0intro:
Applied. Thanks.

Issue #313, opened 2019-12-29
all: deviations from correct standard C in plan9port and required corresponding compiler flags


nsajko:
Plan9port has a whole lot of bugs that stem from the convoluted history of C before and during standardization. Those should all be fixed (I hope nobody disagrees), which will take time.

GCC and Clang provide a few compiler flags for choosing nonstandard variations/dialects of C, so in the mean time we can apply the workaround of adding the needed compiler flags.

`-fno-strict-aliasing` should for some time be passed by 9c to GCC and Clang to disable strict enforcing of standard C aliasing rules. Maybe most programmers do not know about the "strict aliasing" rules mandated by the C standard and the compilers. Quoting the GCC manual (GCC is more lax than the standard):

> In particular, an object of one type is assumed never to reside at the same address as an object of a different type, unless the types are almost the same.  For example, an "unsigned int" can alias an "int", but not a "void*" or a "double".  A character type may alias any other type.

So filling variables of one type using pointers to another type, a quite common pattern, is usually incorrect. See the following links for elaboration: https://blog.regehr.org/archives/959 https://blog.regehr.org/archives/1307 https://web.archive.org/web/20181230041359if_/ http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf

I think this is an example from src/lib9/_p9dialparse.c https://github.com/9fans/plan9port/blob/master/src/lib9/_p9dialparse.c#L39

Another compiler option that should probably be turned on by 9c for some time is `-fno-strict-overflow`. This is similar to the "strict aliasing" situation in that most programmers do not know that signed integers do not wrap around on overflow, rather signed integer overflow is a (programmer) error. Formally it is a class of "undefined behavior", meaning that the compiler (by default) assumes signed integer variables WILL NEVER overflow, and in a situation where a misguided programmer might expect his program to overflow a certain variable, literally anything can happen instead.

This seems to be an example in the hash function in src/cmd/mk/symtab.c: https://github.com/9fans/plan9port/blob/master/src/cmd/mk/symtab.c#L29 . The fix is simple: use an unsigned type for wraparound overflowing.

[...]

PR #314, opened 2019-12-29
9c: enable nonstandard features used by plan9port

nsajko:
The p9p codebase pervasively violates the ANSI and ISO C aliasing rules,
according to which objects may not be accessed through pointers of a
different type, with some exceptions such as (un/signed) char being able
to access any value and a structure and its first member being
interchangeable.
To enable compiler support for the nonstandard aliasing (type punning
wihout unions), the -fno-strict-aliasing flag is needed.

Some aliasing references:
https://blog.regehr.org/archives/959
https://blog.regehr.org/archives/1307
http://dbp-consulting.com/tutorials/StrictAliasing.html
and see the -fstrict-aliasing entry in the GCC manual.

This commit has actually been sitting on the p9p Gerrit since 2016:
https://plan9port-review.googlesource.com/c/plan9/+/1600

Updates #313

### DIFF
--- a/bin/9c
+++ b/bin/9c
@@ -15,6 +15,7 @@ usegcc()
 		-Wno-sign-compare \
 		-Wno-unknown-pragmas \
 		-fno-omit-frame-pointer \
+		-fno-strict-aliasing \
 		-fsigned-char \
 	"
 	# want to put -fno-optimize-sibling-calls here but
@@ -71,8 +72,9 @@ useclang()
 		-Wno-gnu-designator \
 		-Wno-array-bounds \
 		-Wno-unneeded-internal-declaration \
-		-fsigned-char \
 		-fno-caret-diagnostics \
+		-fno-strict-aliasing \
+		-fsigned-char \
 	"
 	cflags="$ngflags -g"
 }


rsc:
> The p9p codebase pervasively violates the ANSI and ISO C aliasing rules,

It's really frustrating to get reports that make claims like this without pointing at a single line of code. I have been trying to avoid violating those aliasing rules. If there are just a few, I'd prefer to clean them up. Where are the pervasive violations?

nsajko:
Well, p9p is not the only C codebase that pervasively violates the ANSI and ISO C aliasing rules, I think maybe even most are like p9p; because compilers started depending on those rules only in the last ten to twenty years?

There is an example in #313 , also the links there to blog posts and the standard draft should be informative.

nsajko:
Btw, the workaround if one wants to alias through incompatible types is to use memcpy or unions.

nsajko:
Almost every time there is a pointer cast, means there is UB near. Just search for `*)`.

nsajko:
https://github.com/9fans/plan9port/blob/7ba9f9467d95fa8d05bb04d36fd4c602e497f529/src/cmd/yacc.c#L959

nsajko:
https://github.com/9fans/plan9port/blob/7ba9f9467d95fa8d05bb04d36fd4c602e497f529/src/cmd/yacc.c#L863

rsc:
For what it's worth, I understand the rules.

It's important to distinguish between the Plan 9 code and the historical >40-year-old Unix code that Plan 9 imported. You really shouldn't draw conclusions about the Plan 9 code as a whole from the 40-year-old stuff. 

I'm still not convinced that yacc fundamentally violates strict aliasing though. It has a tiny memory allocator in it. Obviously you have to be able to write malloc in C, and C compilers don't miscompile malloc. It may be that the use of int instead of char as the base type for the memory pool is technically incorrect, and if you want to change the mem array to be an array of char instead of an array of int, that seems fine to me.

nsajko:
> It's important to distinguish between the Plan 9 code and the historical >40-year-old Unix code that Plan 9 imported. You really shouldn't draw conclusions about the Plan 9 code as a whole from the 40-year-old stuff.

Sorry, I don't know much about that, and I think there may have been a communication error. In particular when I said "p9p is not the only C codebase that pervasively violates the ANSI and ISO C aliasing rules, I think maybe even most are like p9p", I was referring to all codebases in the world.

Regarding malloc, I must say I am not quite sure myself about all the semantics and what is allowed, but I know that in, eg., jemalloc they take much care about not violating aliasing rules and they are themselves not sure what is OK, LOL.

One thing that I am not sure about are pointers to void, more specifically, it may be possible (assuming alignment is OK) to access anything by casting through `void *`. But I am not sure about that.

nsajko:
Casting through coid pointers does not change anything: https://stackoverflow.com/questions/57792531/does-strict-aliasing-occur-when-using-void-pointer-intervening

But often times it is not sure what guarantees regarding correctness do GCC and Clang provide beyond what the standard provides...

rsc:
Strict aliasing matters when you have two different ways to write to a piece of memory at the same time. That's not going on here. The yacc code is clearly running fine. If you have a tool that is complaining about the aliasing in yacc, then let's figure out how to make that tool happy. But let's not just attack every single cast. That's overkill.

nsajko:
@rsc Do you disagree with the following: https://en.cppreference.com/w/c/language/object (search for "effective type").

rsc:
To the extent that the article disallows the implementation of malloc in C, yes.

Again, if you have tools that are rejecting this code or compilers that are miscompiling it, please give the details and let's figure out how to address that. Otherwise let's leave things alone. I'm entirely uninterested in completely hypothetical problems.

dancrossnyc:
It is actually true that the way the standard is written prohibits implementation of parts of the standard library in pure C (e.g., malloc). That said, Russ is very much right: there is the standard, and then there is the real world.

In most places but the C library is written in C and in some places implementations might carefully violate the standard, but in a way that is well-understood by the implementors. This may be technically incorrect (or, more precisely, may technically take advantage of undefined behavior), but it's a calculated risk on the part of people who understand the system at a very deep level.

Plan 9, and by extension the plan9port code, was written in a time and place where the implementors of the system controlled both the OS and the compilers (and in some cases prototyped the hardware, as well). The ported code was carefully moved in ways that, again, were deeply understood by those who did the porting (principally Russ, but also others).

That's not to say that it never has bugs (it does) or can't be improved (it can), but the default should be to assume that the way things are done has been carefully considered and is, for all real-world intents and purposes, correct.

nsajko:
For what it's worth I checked p9dialparse and its callee setport (as part of 9pserve, compiled with default flags, git master) with Ghidra, and could not find any miscompilations. I was looking specifically for elimination of code that the compiler deduced to be dead through type based alias analysis of accesses of sockaddr & co. structs, and I could have missed something so take all of this with more than a grain of salt. I am guessing that GCC did not know the effective type without whole-program analysis, so it could not make any harmful inferences. And you probably don't plan on turning on link time optimization, so those particular cases in lib9 _could_ stay safe.

That said, the yacc code does not need LTO to be misoptimized by the compilers, so there is more reason to fix it (did not check it with Ghidra), not to mention the "if pointers don't fit in integers, there is a problem" comment. Also it does unaligned accesses on the Item pointers in mem0 (not sure how much of a problem that is in practice, but that is also formally UB).

nsajko:
And I meant no insult to Plan 9 at any time, I am well aware that the rules of the game changed after Plan 9 and even maybe to port was already started, and also that Plan 9 had its own dialect of C with a separate compiler. And I am sad that there is not any more of that "whole system creation on all levels" stuff any more.

PR #315, opened 2019-12-29
mk: clean up symtab.c, fix hash

nsajko:
syminit is only called once in main(), and at that point the global
array hash is still all zeroes, thus the call is a no-op; it gets
optimized out.
symdel and symstat also aren't used anywhere, so they are being removed.

The hash function was abstracted out of symlook and symdel, and changed
so as not to use signed integer overflow because that is forbidden in
standard C.

A version of this commit has actually been sitting on the p9p Gerrit
since 2016:
https://plan9port-review.googlesource.com/c/plan9/+/1572

Updates #313

Change-Id: I5478ae03f208a92a6ba16060486b946dcc4b2cf1

### DIFF
--- a/src/cmd/mk/symtab.c
+++ b/src/cmd/mk/symtab.c
@@ -1,7 +1,7 @@
 #include	"mk.h"
 
 #define	NHASH	4099
-#define	HASHMUL	79L	/* this is a good value */
+#define	HASHMUL	79UL	/* this is a good value */
 static Symtab *hash[NHASH];
 
 void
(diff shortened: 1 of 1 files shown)


rsc:
Again, bug fixes and fixes for modern too-clever compilers are fine.
Please avoid "clean up" fixes.
All this PR needs to do is change "long h" to "ulong h" in a few places.

rsc:
Thanks for updating the code.
Is the (ulong) cast necessary? I don't see why.

rsc:
Thanks. The HASHMUL redefinition could have been left out too, but I took it.

PR #317, opened 2019-12-31
lib9/fmt: fix data race on fmtalloc

nsajko:
The spin lock with volatile was not functional, and it did not protect
fmtalloc.nfmt anyway.

Change-Id: Ib4e8013f4f730d57f88d98c11b5bac4a11b20ba7

### DIFF
--- a/src/lib9/fmt/fmt.c
+++ b/src/lib9/fmt/fmt.c
@@ -14,7 +14,7 @@ typedef struct Convfmt Convfmt;
 struct Convfmt
 {
 	int	c;
-	volatile	Fmts	fmt;	/* for spin lock in fmtfmt; avoids race due to write order */
+	Fmts	fmt;
 };
 
 static struct
@@ -117,16 +117,16 @@ fmtfmt(int c)
 {
 	Convfmt *p, *ep;
 
+	__fmtlock();
 	ep = &fmtalloc.fmt[fmtalloc.nfmt];
 	for(p=fmtalloc.fmt; p<ep; p++)
 		if(p->c == c){
-			while(p->fmt == nil)	/* loop until value is updated */
-				;
-			return p->fmt;
+			Fmts tF = p->fmt;
+			__fmtunlock();
+			return tF;
 		}
 
 	/* is this a predefined format char? */
-	__fmtlock();
 	for(p=knownfmt; p->c; p++)
 		if(p->c == c){
 			__fmtinstall(p->c, p->fmt);


dancrossnyc:
While I'm fairly certain that the original code was, in fact, incorrect, I'm not certain that the original code is incorrect. Did you see a failure?

In particular, reasoning through the logic in `__fmtinstall`, I don't see a case where a thread can observe that p->c == c while p->fmt is nil, and the volatility of `fmt` should ensure that. But I'm willing to be convinced that I'm wrong.

That said, this change does seem clearer and I'm having a hard time believing that that lock is often contended.

rsc:
Please make an effort to use identifiers that fit in with the surrounding code. The name `tF` is unlike any other name in this code base - there is no mid-word capitalization. Same thing happened with the `buildEnvironmentVariables` file. 

If you want to help us clean things up, I have no problem with that, as long as they are things that really need cleaning up (this one probably qualifies). But please try to make your changes look like the surrounding code, so that no one can tell you made changes just from the way things look. 

I realize this is difficult to pick up, and we're happy to keep pointing these things out. Just please be extra aware and look closely at the surrounding context.

Thanks.

rsc:
In this case, you don't need a variable at all. Just move the __fmtlock up, and change the while loop to __fmtunlock, leaving the "return p->fmt" after the unlock, same as below. It is fine to read from p->fmt while not holding the lock because once we see p->c is set while holding the lock we know p->fmt has been written and will never been updated again.

nsajko:
I will try for things like `tF` not to happen again, sorry.

rsc:
I'd like to fix this a different way, that preserves the fast path. I'll commit a change and mark it fixes #317.

dancrossnyc:
Then we should close this PR. However, is the fast path actually all that much faster?

It looks like the issue here is that evaluating formatters races with installing a new formatter. However, format installation is infrequent at best, and output in multiple threads hitting this code path is likely to be minimal (I'd imagine that the dominant cost would be the underlying IO mechanism). Thus, I'm wondering if this lock is ever seriously contended? Please note that this is a question, not a statement of fact, and of course I have no evidence either way, but if such evidence exists, I'd love to see it.

rsc:
Fixed in 7ba9f946.

rsc:
@dancrossnyc, the code goes out of its way to avoid the lock on the fast path and it seems worth preserving that decision. If you've got threads doing nothing but formatting, it seems pretty suboptimal to make them fight over a lock acquired and released once per % sign in the formats. I'd much rather keep concurrent formatting conflict-free, so I checked in a change to do rlock/runlock for that path.

dancrossnyc:
I'm afraid that the rwlock change actually cases a performance regression in the fast path. I did a benchmark and compared against a simple mutex; in all cases the reader/writer lock is slower. I'll send you my results.

PR #318, opened 2019-12-31
mk: fix out of bounds access

nsajko:
Updates #313

Change-Id: I0e27e15feacb77391bc1decee7cf720d64d14586

### DIFF
--- a/src/cmd/mk/archive.c
+++ b/src/cmd/mk/archive.c
@@ -1,6 +1,6 @@
 #include	"mk.h"
 #define	ARMAG	"!<arch>\n"
-#define	SARMAG	8
+#define	SARMAG	(sizeof(ARMAG) - sizeof(""))
 
 #define	ARFMAG	"`\n"
 #define SARNAME	16
@@ -102,7 +102,7 @@ atouch(char *name)
 		LSEEK(fd, SARMAG, 0);
 		while(read(fd, (char *)&h, sizeof(h)) == sizeof(h)){
 			for(i = SARNAME-1; i > 0 && h.name[i] == ' '; i--)
-					;
+				;
 			h.name[i+1]=0;
 			if(strcmp(member, h.name) == 0){
 				t = SARNAME-sizeof(h);	/* ughgghh */
@@ -116,6 +116,18 @@ atouch(char *name)
 		}
 	}
 	close(fd);
+}
+
+static int
+allspaces(char *a, int n)
+{
+	int i;
+	for (i = 0; i < n; i++) {
+		if (a[i] != ' ') {
+			return 0;
+		}
+	}
+	return 1;
 }
 
 static void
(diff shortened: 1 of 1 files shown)


src/cmd/mk/archive.c
```diff
@@ -138,25 +140,25 @@ atimes(char *ar)
...
 			if(readn(fd, name, namelen) != namelen)
 				break;
 			name[namelen] = 0;
-		}else if(memcmp(h.name, "// ", 2) == 0){ /* GNU */
+		}else if(memcmp(h.p.name, "// ", 3) == 0){ /* GNU */
 			/* date, uid, gid, mode all ' ' */
-			for(i=2; i<16+12+6+6+8; i++)
-				if(h.name[i] != ' ')
+			for(i=2; i < SARNAME+12+6+6+8; i++)
+				if(h.h[i] != ' ')
```
dancrossnyc: The crux of this change seems to be this loop, but the fix seems oddly disproportionate to the offense committed; in particular, something like the following feels a little easier to negotiate:

```
static int
allspaces(char* buf, int len)
{
        for(int i=0; i<len; i++)
                if(buf[i] != ' ')
                        return 0;
        return 1;
}

                // and at the call site....
                if(!allspaces(h.name, sizeof(a.name)) ||
                    !allspaces(h.date, sizeof(h.date)) ||
                    !allspaces(h.uid, sizeof(h.uid)) ||
                    !allspaces(h.gid, sizeof(h.gid)) ||
                    !allspaces(h.mode, sizeof(h.mode)))
                        goto skip;
```

nsajko: @dancrossnyc done.

src/cmd/mk/archive.c
```diff
@@ -151,11 +165,12 @@ atimes(char *ar)
 			if(readn(fd, name, namelen) != namelen)
 				break;
 			name[namelen] = 0;
-		}else if(memcmp(h.name, "// ", 2) == 0){ /* GNU */
+		}else if(memcmp(h.name, "// ", 3) == 0){ /* GNU */
 			/* date, uid, gid, mode all ' ' */
-			for(i=2; i<16+12+6+6+8; i++)
-				if(h.name[i] != ' ')
-					goto skip;
+			if(!allspaces(&h.name[3], sizeof(h.name) - 3) || !ALLSPACES(h.date) ||
```
dancrossnyc: This wraps a little oddly, but otherwise ok. I'm not so thrilled about the macro, but understand why you are using it.

nsajko: @dancrossnyc If I may ask, what is wrong with the macro? Are you afraid somebody will use it with a pointer instead of with an array?

nsajko:
@dancrossnyc the commit message is fixed.

src/cmd/mk/archive.c
dancrossnyc: That's a possibility, but it's more an aesthetic dislike of macros in general. As a recovering Lisp hacker, I try to avoid the preprocessor wherever I can...including macros. But I also understand this is subjective.

That said, adding something like a macro has a cost; indeed, this is true of adding any code. One must always weigh whether incurring that cost is sufficiently rewarded by additional benefit along some important dimension: readability, maintainability, reusability, performance, etc. In this case, I think the function is worth it because the alternative is a set of clunky repetitive loops or a more involved change like your original union. Here, I don't think the macro is worth it because it saves a single parameter from a handful of function calls in a single boolean expression. I doubt it will ever be misused, but I don't think it pulls its weight enough to be worth it in the first place. But as I said, this opinion is filtered through the lens of my own biases, and is inherently subjective.

To the extent that I have a more defensible argument, it's that it doesn't really fit with the style of plan9ports, and plan9 in general: yes, plan9 _does_ use function-style macros occasionally, but usually as a substitution for inline functions and rarely for parameter hiding in this manner.

nsajko: @dancrossnyc I removed the macro from the commit.

nsajko: Thanks for the elaboration on macros, BTW.

dancrossnyc: Sure thing!

PR #319, opened 2020-01-01
all: explicitly convert type to prevent signed overflow

nsajko:
The problem is usually that unsigned char-s get unintuitively promoted
to signed int-s implicitly.

Signed integer overflow is an error in (today's) C.

Updates #313

Change-Id: I87ebad5dd21991eb164492fce586b115d591ad23


src/libthread/thread.c
```diff
@@ -87,8 +87,7 @@ threadstart(uint y, uint x)
 	ulong z;
 
 //print("threadstart\n");
-	z = (ulong)x << 16;	/* hide undefined 32-bit shift from 32-bit compilers */
```
nsajko: @0intro I am not sure if I should have joined these two lines. Do you know their original intention? I assume that back on Plan9 `ulong`s were 32-bit? But I think that for plan9port they are always 64-bit?

0intro: ulong are always 4-byte on Plan 9, but are either 4-byte (on 32-bit platforms) or 8-byte (on 64-bit platforms) where plan9port runs (Unix). I think this code was added to handle these two cases.

nsajko: @0intro I removed this particular change from the commit (it wasn't connected to the rest of the commit anyway), and added more fixes.

nsajko:
Fixed the conflicts.

rsc:
I don't understand some of the changes in this patch. Why are there <<8 and <<16 that are being edited too?

nsajko:
The `byte<<8` and `byte<<16` don't need casts, I just put them there for consistency within the bitwise "or" operation (but then gave up on consistency after I got tired of all that typing). Do you want casts only where strictly necessary?

rsc:
> Do you want casts only where strictly necessary?

Mostly I want minimal diffs. So if the casts are not there then don't add them unless necessary.
If the casts are there already in the main repo, then don't remove them.
Thanks.

nsajko:
This should be OK now.

PR #323, opened 2020-01-07
cmd/venti/srv: split memory allocation call

nsajko:
This splits a certain vtmallocz call in mkihash into two vtmallocz
calls. The first issue this fixes is that the C aliasing rules were not
respected in the code before this commit. The other thing is that this
enables better memory alignment guarantees.

Updates #313

Change-Id: Ia4f3e0fc85facc778193f5e977d4f99a1a9abd23

### DIFF
--- a/src/cmd/venti/srv/icache.c
+++ b/src/cmd/venti/srv/icache.c
@@ -68,8 +68,8 @@ mkihash(int size1)
 		size <<= 1;
 	}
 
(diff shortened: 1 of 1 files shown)


rsc:
FWIW the commit message form here is not exactly the same as in Go.
It's just "program:" or "library:" not "full/path/from/src:", so here instead
of "cmd/vent/srv:" the next time just use "venti:". Thanks.

PR #327, opened 2020-01-08
acme: exclude terminating null from strncpy bound

nsajko:
Change-Id: I9d5ad1c5821d633c81a322841143164364034d97

### DIFF
--- a/src/cmd/acme/acme.c
+++ b/src/cmd/acme/acme.c
@@ -767,7 +767,8 @@ waitthread(void *v)
 				if(strncmp(w->msg, "libthread", 9) != 0){
 					p = emalloc(sizeof(Pid));
 					p->pid = pid;
-					strncpy(p->msg, w->msg, sizeof(p->msg));
+					strncpy(p->msg, w->msg, sizeof(p->msg) - 1);
+					p->msg[sizeof(p->msg) - 1] = 0;
 					p->next = pids;
 					pids = p;
 				}


rsc:
How are you finding these? Are you running some kind of sanitizer that is keeping you from using the code?

There doesn't seem to be much benefit to most of these changes, and I worry about the amount of churn accidentally introducing bugs that users might actually hit.

Thanks.

nsajko:
@rsc GCC pointed out with a warning that this change would be a good idea. Most of the benefit for this particular change is just getting rid of that GCC warning.

That said, I really feel bad about your second paragraph, because it seems to refer to changes like fixing signed integer overflow or fixing aliasing violations, which are actual important bugfixes (especially while the non-standard compiler flags are not used, see #313) .

> Are you running some kind of sanitizer that is keeping you from using the code?

Yes, I run Google's sanitizers (compiled with GCC), and I aim to fix defects that break aliasing rules, signed integer overflows, data races, make signals be handled safely. I hope you are not suggesting this is harmful to plan9port.

rsc:
I agree about fixing problems that cause compilers to miscompile code,
such as signed integer overflow assumptions.
I am less concerned about compilers making style judgements.

And as I said, I am very concerned about changes that have little positive
impact having unintended negative impact. I've seen that happen over and
over again when people set out to "fix" newly reported problems found by
over-aggressive analysis or fuzzing. Too often the result is new bugs worse
than the old ones. For the specific case of strncpy, I would rather that calls
be convered to use strecpy or utfecpy, which handle termination properly
in a single line instead of requiring a second. But of course some calls may
be depending on the zeroing of the rest of the buffer, which strncpy does
and the others don't. So it is important to make each change carefully.

Issue #26, opened 2015-03-31
Can have spaces in file or directory name


blakemcbride:
There is no way to select a directory that has a space in its name, e.g. "MY DIR".  Same for files with spaces in them.  This is the single biggest reason I don't use acme all of the time.

gleydsonsoares:
well, just use $PLAN9/bin/B

eg.:
B 'MY DIR' or B 'FILE NAME'

* Requires plumber(4) running.

rbns:
while this enables opening the files as soon as the window with that file is selected, everything after the space is removed. thus, the file can't be `Put` etc.

maybe we can patch acme so that it understands the behaviour os `ls` from plan9port which wraps paths with spaces in `'`

i can try to look into this, but i don't really know my way round in the acme code. i guess it can't be _too_ hard for someone who understands whats happening

blakemcbride:
I like acme a lot.  Pike made a lot of really good choices that make it a
remarkable editor.  I would use it if I could, but I do not.  I don't
because of a handful of minor things that are major impediments to its
use.  I call them minor things because I believe that they are really easy
to correct.  I don't do it myself because I literally work every moment I
am not spending with what little time I spend with my family.  It is not
possible for me to correct.

I have a very small list of, what I consider, major issues (that are minor
issues to correct).  The number one issue is not supporting spaces between
directory and file names.  Every time I try to make major use of acme I get
hit by this.

IMO, it is also sad that Pike's ideas have become to some a religion.
While Pike has contributed greatly to the field, he's not infallible.  A
few tweaks / corrections to his masterpiece could make all the difference
to the general use and applicability of his ideas.

Just an opinion.

Blake McBride

blakemcbride:
Still can't edit files with spaces in their name or their path.  Also, can't select fonts with spaces e.g. "Courier New".

jxy:
It may not be difficult to use the library functions in quote(3) to perform quote/unquote.  The result may be trivially backward compatible.  @blakemcbride, since you comment on this issue every year, perhaps you can try your hand at it?

blakemcbride:
I check it every so often to see if it now works (or if p9p has disappeared altogether).

dertuxmalwieder:
Adding myself to CC for the time being. I might try to work on this after my stack of other TODOs has been processed. (Which might be in 2020 or so.) Anyone else?

henesy:
[...]

PR #329, opened 2020-01-10
ACME: Support editing file names with spaces

farhaven:
Hi,

this PR is a fix for #26. It works mainly by tweaking the logic that decides which part of the tag of a window is the file name:

Everything up to the first tab character in the tag is now considered a file name

Visually, the difference is relatively minor. It takes a bit getting used to the blank space between the file name and the `Del`, but apart from that it seems to do the job.

I've been using this patch for a while and haven't observed negative side effects yet.

### DIFF
--- a/man/man1/acme.1
+++ b/man/man1/acme.1
@@ -107,6 +107,8 @@ The tag contains a number of
 blank-separated words, followed by a vertical bar character, followed by anything.
 The first word is the name of the window, typically the name of the associated
 file or directory, and the other words are commands available in that window.
+The file name is separated from all other words in the tag by a tabulator character
+to allow editing of files with spaces in their name or path.
 Any text may be added after the bar; examples are strings to search for or
 commands to execute in that window.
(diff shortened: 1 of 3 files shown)


farhaven:
And just as luck would have it, I've just now discovered that it behaves funny if you remove the tab. I'll amend the patch.

blakemcbride:
Thanks!  This was the #1 reason I couldn't use ACME.

farhaven:
Good to hear. Keep in mind though that some tools, for example `acme-lsp` have trouble extracting the file name from tags if this patch is applied (they only consider ASCII spaces to separate words, not tabs).

fhs:
Note: there is another PR open that adds support for spaces in filenames: https://github.com/9fans/plan9port/pull/104

rsc:
Sorry but no. Rob and I have been discussing this. Eventually we will make a change, but we're not soliciting external changes for this. There are lots of subtle implications.

PR #330, opened 2020-01-13
cmd/acme: be more careful with pipes in X commands

### LINKED ISSUE #9: cmd/acme: Edit X/.../ broken

jacobvosmaer:
This is an alternative to https://github.com/9fans/plan9port/pull/222 . The goal is to fix #9 and #219: Acme crashes when an `X` or `Y` edit visits a window different from where the command was written, and if the loop body invokes a pipe (`<`, `>` or `|`).

### DIFF
--- a/src/cmd/acme/ecmd.c
+++ b/src/cmd/acme/ecmd.c
@@ -579,12 +579,14 @@ x_cmd(Text *t, Cmd *cp)
 	return TRUE;
 }
 
+static Window *Xorigwin = nil;
+
 int
 X_cmd(Text *t, Cmd *cp)
 {
-	USED(t);
-
+	Xorigwin = t->w;
 	filelooper(cp, cp->cmdc=='X');
+	Xorigwin = nil;
 	return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


src/cmd/acme/ecmd.c
```diff
@@ -626,8 +628,12 @@ runpipe(Text *t, int cmd, Rune *cr, int ncr, int state)
...
 	run(w, runetobyte(s, n), dir.r, dir.nr, TRUE, nil, nil, TRUE);
 	free(s);
-	if(t!=nil && t->w!=nil)
-		winunlock(t->w);
+	/*
+	 * Only winunlock if we are in the window the command originated from.
+	 * With X and Y this is not automatically the case.
+	 */
+	if(t!=nil && t->w!=nil && (Xorigwin==nil || Xorigwin==t->w))
+			winunlock(t->w);
```
jacobvosmaer: Oops one tab too many

rsc:
I don't think this is right either. I think the lock really does need to be released around the pipe execution.

jacobvosmaer:
I don't really understand what I'm doing but I wonder if the behavior of this PR is unclear.

@rsc as I understand it this change keeps the same `winunlock` behavior we had before, with the exception that during X (`Xorigwin!=nil`), we iterate over windows some of which weren't locked to begin with. So while `Xorigwin!=nil` we should _only_ unlock if the the window the current text is in is the "originating window" (`Xorigwin==t->w`).

Sorry if all that was obvious.

Are you saying we should `winlock` all windows that X visits before we run commands on them?

jacobvosmaer:
@rsc is 1ad3ad52 then what you had in mind?

PR #331, opened 2020-01-14
acme: allow spaces in window names

### LINKED ISSUE #26: Can have spaces in file or directory name

rsc:
There are many things we could do to make this work.
an environment variable to control the character.
Another option would be to use U+00A0 (non-breaking space),
which renders the same as space.

This change avoids changing the separator character and instead
assumes that if the left side of the tag already ends in " Del Snarf |"
then what comes before that is the file name.
Acme already aggressively preserves the "Del Snarf |",
so this should work decently well as a stop-gap.
We can always try something else later.

Fixes #26.
Fixes #104.
Fixes #329.

### DIFF
--- a/src/cmd/acme/look.c
+++ b/src/cmd/acme/look.c
@@ -610,7 +610,7 @@ expandfile(Text *t, uint q0, uint q1, Expand *e)
 	if(nname == -1)
 		nname = n;
 	for(i=0; i<nname; i++)
-		if(!isfilec(r[i]))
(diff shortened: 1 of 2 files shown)


rsc:
This is for discussion, not yet for submission.

rsc:
Merged with some fixes.

PR #337, opened 2020-01-14
tar: remove /bin/ when invoking decompressors

phonologus:
…essors.

### DIFF
--- a/src/cmd/tar.c
+++ b/src/cmd/tar.c
@@ -179,7 +179,6 @@ static int
 push(int fd, char *cmd, int input, Pushstate *ps)
 {
 	int nfd, pifds[2];
-	String *s;
 
 	ps->open = 0;
 	ps->fd = fd;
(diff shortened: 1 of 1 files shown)


dancrossnyc:
Hmm. Was this causing a specific problem?

Also, a meta-note: the commit message should start with a line that just says, `tar:` not the path to the source file.

rsc:
@dancrossnyc, if the code is fine then it's fine to just reword the commit message during squash and merge.

rsc:
I think the specific problem is that if tar runs /bin/gunzip it won't find it.

phonologus:
Apologies for the formatting. This was my first ever contribution on
GitHub: still learning.

Specific problem is that tar is calling /bin/bzip2, /bin/gzip (and
/bin/compress), and not the plan9port-specific ones.

S.

--
Sean Jensen

rsc:
It doesn't need to call plan9-specific ones, it just needs to call ones that exist. And they are never in /bin on Unix. Using the path (which you'd get without the /bin/) is the right behavior.

phonologus:
Yes, thanks Russ.

Fedora, which I use, actually does put gzip and bzip2 into /bin... 

Fedora has no (un)compress by default: that's how I discovered the problem, as I use compress a lot.

dancrossnyc:
@rsc right, it wasn't clear to me that this was actually triggering in practice, but clearly it is.

The change as-is seems reasonable.

PR #341, opened 2020-01-15
9pfuse: update errortab

lego12239:


dancrossnyc:
These seem to be everywhere and are defined by POSIX, so it seems ok.

PR #346, opened 2020-01-15
winwatch: Plan 9-ify.

dancrossnyc:
This is new code, and custom to plan9port.  Make it
conform more closely to plan9 style.


markvanatten:
Thanks!

PR #350, opened 2020-01-16
libthread: remove adjustments to stack pointer/size

### LINKED ISSUE #109: [RFC] segfault: libthread/thread.c breaks 16-byte aligned stack
**context**
_linux/gcc/musl/x86_64_

GCC is apparently warranted to suppose the stack is aligned to 16 bytes and when provoked (-O2) will use an obscure x86 obstruction called MOVAPS to copy 16 bytes at once to memory from a 16 byte register such as %xmm0. If the destination operand, say (%rbp), is not aligned to 16 bytes this instruction triggers a segmentation fault.

**proposed solution**
_src/libthread/thread.c/threadalloc:136_

`/* leave a few words open on both ends */`
`t->context.uc.uc_stack.ss_sp = (void*)(t->stk+8);`

It's tempting to just drop the open word at the beginning:
`t->context.uc.uc_stack.ss_sp = (void*)(t->stk);`

But those words were obviously left open with a reason so I went with:
`t->context.uc.uc_stack.ss_sp = (void*)(t->stk+16);`

This seems to solve the problem in my case.

Does anyone know why these words were left open or why you can assume you have extra memory available to begin with?

dancrossnyc:
When setting the stack pointer and size in the
ucontext structure when creating a thread, the
stack pointer was offset by 8 bytes and the size
was 64 less than allocated.  This appears to be
debugging code from some time ago, but I couldn't
quite figure out what it's purpose was (perhaps
poisoning the top/bottom of the stack to detect
under/overflow?).  The 8-byte offset for the
stack pointer, in particular, was causing instructions
like `MOVAPS` to fail on some platforms: the ABI
assums 16-byte stack alignment on entry to a
function.

Anyway, remove the adjustments.

Fixes #109

### DIFF
--- a/src/libthread/thread.c
+++ b/src/libthread/thread.c
@@ -137,9 +137,8 @@ threadalloc(void (*fn)(void*), void *arg, uint stack)
 //print("makecontext sp=%p t=%p startfn=%p\n", (char*)t->stk+t->stksize, t, t->startfn);
 
 	/* call makecontext to do the real work. */
-	/* leave a few words open on both ends */
-	t->context.uc.uc_stack.ss_sp = (void*)(t->stk+8);
-	t->context.uc.uc_stack.ss_size = t->stksize-64;
+	t->context.uc.uc_stack.ss_sp = t->stk;
+	t->context.uc.uc_stack.ss_size = t->stksize;
 #if defined(__sun__) && !defined(__MAKECONTEXT_V2_SOURCE)		/* sigh */
 	/* can avoid this with __MAKECONTEXT_V2_SOURCE but only on SunOS 5.9 */
 	t->context.uc.uc_stack.ss_sp =


src/libthread/thread.c
```diff
@@ -138,8 +138,8 @@ threadalloc(void (*fn)(void*), void *arg, uint stack)
 
 	/* call makecontext to do the real work. */
 	/* leave a few words open on both ends */
```
mirtchovski: this comment is now incorrect, isn't it?

dancrossnyc:
It is, indeed.

Good catch; fixed.

mirtchovski:
Assuming this was never used anywhere this LGTM (but still wait for @rsc). Would another option be to change `(void*)(t->stk+8)` to `(void*)(t->stk+16)`, just in case this is still useful somewhere?

dancrossnyc:
Yeah, potentially, but I can't see why it's necessary. I traced through the git history and near as I can tell, this is (or perhaps more correctly was) for debugging at some point, but is otherwise not used. The constants involved are somewhat magic and unexplained.

But maybe it serves some purpose I am unaware of; hence sending a pull request and soliciting a review from @rsc instead of just pushing directly.

rsc:
I'd rather not do this. If this change fixes anything, then it is only by accident.
Nothing should depend on exactly what alignment that pointer is.
Anything that wants alignment should do the work itself.
See for example Darwin-386.c and Darwin-x86_64-swapcontext.c.

dancrossnyc:
Do you mean that e.g. `makecontext` should make sure that it's properly aligned? That doesn't seem to be happening in the referenced problem report. What's the point of reserving those bytes at the beginning and end, anyway? It seems like it's just inviting problems for no particularly good reason.

rsc:
It's not inviting any problems. It's turning simple off-by-ones into non-problems.

dancrossnyc:
If we're relying on the ucontext functions to handle rounding the stack pointer correctly, then shouldn't we also expect them to handle those off-by-one errors you're mentioning correctly as well?

Regardless, perhaps reducing the stack size may prevent some kind of weirdness, or maybe the stupidity of OpenBSD's MAP_STACK (I just talked to Austin about that) but as per issue #109, it's also creating a problem on at least one platform. If we want to keep some space at the bottom of the stack available, then fine I guess, but surely we can adjust the offset to be 16 instead of 8.

rsc:
Sorry, but this is the wrong fix for #109.

dancrossnyc:
I'm sorry, Russ, but that response begs the question: what is the right fix?

If relying on the *context() functions to get this right, I'll note that in the current design of libthread, this doesn't seem to be handled correctly pretty much anywhere where it provides the definitions of those functions (e.g., src/libthread/OpenBSD-x86_64.c and src/libthread/OpenBSD-x86_64-asm.S; ironically, the former uses MOVAPS, the instruction that kicked all of this off). Should we go through and change the definitions of `setcontext` etc for various platforms to ensure alignment?

rsc:
Like I said [above](https://github.com/9fans/plan9port/pull/350#issuecomment-575657404), the fix is to change makecontext on the systems with alignment requirements to respect those requirements.

You do not need to change setcontext or getcontext. Those do not construct new contexts from whole cloth.

dancrossnyc:
There's really no need to put in that way: this actually is the first time you mentioned `makecontext`, though I inferred it before.

I'm happy to modify `makecontext` though of course `setcontext` is where the value is actually consumed and one would assume it to be the logical place to validate whatever system-mandated requirements are relevant.

These constants are, in some ways, a separate issue: I looked up their origin in the git history, and these values are undocumented. I argue that they should go, but if you want to keep them, they should at least be documented in some reasonable way: either in a comment or in a commit message. Perhaps that will short-circuit future discussions of this nature.

rsc:
Apologies. Happy to talk in person about this next week.

dancrossnyc:
No worries. Sure, let's chat next week.

rsc:
Submitted 4197af41 to explain stack border a bit more, per our discussion.

PR #354, opened 2020-01-19
libthread: rm unused ARM (get|set)mcontext

nsajko:
They were just a duplicate of my(get|set)mcontext from the other
assembly file, and unused from threadimpl.h.

Change-Id: Id8003e5177ed9d37a7f0210037acbe55bbf7f708


dancrossnyc:
No. Libthread is really subtle code and Russ and I are going to talk about it next week. Please don't start deleting things from it.

nsajko:
OK

rsc:
There are subtle things about libthread, but this is not one of them.
The only possible way this code could be relevant is if compiled on FreeBSD 4 arm systems,
and even if those existed it would result in a macro redefinition error first
(two different #defines for setcontext).
This PR is OK.

PR #358, opened 2020-01-26
wintext: replace bash with rc

mk-f:
``"``, ``""`` and ``wintext`` will fail if ``/bin/bash`` is not present, replace wintext with rc.

### DIFF
--- a/bin/wintext
+++ b/bin/wintext
@@ -1,22 +1,14 @@
-#!/bin/bash
+#!/usr/local/plan9/bin/rc
 
-case "$winid" in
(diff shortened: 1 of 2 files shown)


nicolagi:
Looks like this was done in these two commits https://github.com/9fans/plan9port/commit/0bd14783426d137428ffae7cd89cfc06b88d1b11 https://github.com/9fans/plan9port/commit/6c8e44dd2e3b346ad5e313830ef22ee7b0a9df04 so the PR could be closed?

PR #365, opened 2020-02-11
acme: ^S -> Putall

dexen:


jroimartin:
Why to keep adding keybindings? I like the simplicity of having just a few keybindings for text edition. The GUI and the mouse are always there for other actions.

jroimartin:
Actually, @rsc what was the reason for adding Kcmd+{c,x,v,z,Z}? I guess they can be convenient when editing from a laptop without a proper mouse, but {Alt, Cmd} + trackpad seem to work quite well also for those actions.

dexen:
> Why to keep adding keybindings?

i'm very happy with mouse for all the actions - including Put, and Putall - when editing code
however writing, or redacting, free flowing text seems to be helped with different, more immersive mode

rsc:
Sorry but we're not adding any more keybindings, especially not for Control sequences.
The Kcmd actions were added to make editing nicer on single-button Mac trackpads but I think that's where the line stops. Maybe they were a mistake, but so be it.

Note that we can't even do Control equivalents on non-Mac because ^C too often needs to be interrupt.

PR #366, opened 2020-02-14
netfiles: use ssh(1) for network file access

jxy:
With ControlMaster configured, ssh(1) alone works fine for network file access.

### DIFF
--- a/src/cmd/netfiles/main.c
+++ b/src/cmd/netfiles/main.c
@@ -324,7 +324,7 @@ filethread(void *v)
 	threadsetname("file %s", a->file);
 	w = newwin();
 	winname(w, a->file);
-	winprint(w, "tag", "Get Put Look ");
+	winprint(w, "tag", "Get Put ");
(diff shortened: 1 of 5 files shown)


src/cmd/netfiles/main.c
```diff
@@ -324,7 +324,7 @@ filethread(void *v)
 	threadsetname("file %s", a->file);
 	w = newwin();
 	winname(w, a->file);
-	winprint(w, "tag", "Get Put Look ");
+	winprint(w, "tag", "Get Put ");
```
dancrossnyc, line 327: Is this change deliberate?

jxy: It's an accident.  Can you remove it from this PR?  I can squash it when I get around to it.

dancrossnyc: Not easily, no, because I'm out of town.

jxy:
Cleaned up and removed the change in `main.c`.

PR #371, opened 2020-03-06
acme: add noscroll-alike behaviour to +Errors windows

### LINKED ISSUE #370: acme: add noscroll-alike behaviour to +Errors windows
Currently any output to +Errors causes the view to jump and scroll all the way to the bottom. This makes it impossible to go through the output without waiting for the process to finish, and that is counter-productive.

The problem was solved for _win_ windows: first by the Noscroll command, and then by making the scroll  conditional and only jump to the bottom if the window already showing the output point. 

Let's consider adopting same technique for the +Errors windows.

yrk-lab:
Fixes #370.

### DIFF
--- a/src/cmd/acme/xfid.c
+++ b/src/cmd/acme/xfid.c
@@ -407,8 +407,6 @@ xfidread(Xfid *x)
 static int
 shouldscroll(Text *t, uint q0, int qid)
 {
-	if(qid == Qcons)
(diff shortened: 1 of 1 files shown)


yarikk:
After using this for a while I realised it's better to  have them auto-scrolling than having trouble to find where my output go.

Issue #372, opened 2020-03-08
fontsrv X11: incorrect baseline


dzklaim:
Apparently, the baseline of fonts is incorrect when using X11. This leads to clipped letters like in #18 

Changing src/cmd/fontsrv/x11.c:81 to match f->height scaling seems to solve the problem (also "because" is misspelled).

`f->originy = face->descender * 1.35; // bbox.yMin (or descender)  is negative, because the baseline is y-coord 0`

markvanatten:
Could you turn this into a pull request?

dzklaim:
done!

markvanatten:
Thanks!

rsc:
Applied.

PR #373, opened 2020-03-12
scale f->originity to match f->height

dzklaim:
Hopefully this fixes the incorrect baseline of fonts in X11 described in #372. The number 1.35 was empirically determined in #111. There should be a better approach ...


lego12239:
This helps me with DejaVuSerif https://github.com/9fans/plan9port/issues/375 .
Thanks!

gdiazlo:
It worked for me too, thanks.

PR #380, opened 2020-03-23
cmd/rio/winwatch: let initdraw() pick the font if none passed with -f

nicolagi:
This way winwatch will honor the $font environment variable.


markvanatten:
Thanks!

Could you update the man page accordingly and submit that too?

nicolagi:
Good idea. Done.

PR #381, opened 2020-03-23
acme: add env var for color scheme configuration

lego12239:
acmecscheme environment variable is used for quick change of
color scheme without acme recompiling.

acmecscheme contains a color scheme specification in the next
format:

ENTITY1=COLOR1.1:COLOR1.2:...;ENTITY2=COLOR2.1:COLOR2.2:...

Where entities is one of "tag" or "text" (without quotes); and
colors are specified by rgba code. color code can be missed and
default is used in this case.

E.g.:

acmecscheme="tag=0x454545ff:::0xd3d3d3ff:;text=0x1a1a1aff:::0xc3c3c3ff:" acme

### DIFF
--- a/src/cmd/acme/acme.c
+++ b/src/cmd/acme/acme.c
@@ -31,6 +31,12 @@ int		snarffd = -1;
 int		mainpid;
 int		swapscrollbuttons = FALSE;
 char		*mtpt;
+int		tag_col_back = 0xeaffffff, tag_col_high = 0x9eeeeeff,
+		tag_col_bord = 0x8888ccfd, tag_col_text = 0x000000ff,
+		tag_col_htext = 0x000000ff,
+		text_col_back = 0xffffeaff, text_col_high = 0xeeee9eff,
+		text_col_bord = 0x99994cff, text_col_text = 0x000000ff,
+		text_col_htext = 0x000000ff;
 
 enum{
 	NSnarf = 1000	/* less than 1024, I/O buffer size */
@@ -49,6 +55,7 @@ void	shutdownthread(void*);
 void	acmeerrorinit(void);
 void	readfile(Column*, char*);
 static int	shutdown(void*, char*);
+static void set_color_scheme(char*);
 
 void
 derror(Display *d, char *errorstr)
@@ -137,6 +144,11 @@ threadmain(int argc, char *argv[])
 	acmeshell = getenv("acmeshell");
 	if(acmeshell && *acmeshell == '\0')
 		acmeshell = nil;
+	p = getenv("acmecscheme");
+	if(p != nil){
+		set_color_scheme(p);
+		free(p);
+	}
 	p = getenv("tabstop");
 	if(p != nil){
 		maxtab = strtoul(p, nil, 0);
@@ -363,6 +375,63 @@ shutdownthread(void *v)
 		shutdown(nil, msg);
 }
 */
+
+static char* set_color_scheme_for_entity(int **ecols, char *csspec);
+
+/*
+ * Set acme color scheme from a textual specification csspec. spec format:
+ * ENTITY1=COLOR1.1:COLOR1.2:...;ENTITY2=COLOR2.1:COLOR2.2:...
+ * Any color can be empty(e.g. ENTITY1=COLOR1.1::COLOR1.3...).
+ */
+static void
+set_color_scheme(char *csspec)
+{
+	char *p;
+	int *tagc[] = {&tag_col_back, &tag_col_high, &tag_col_bord, &tag_col_text,
+	              &tag_col_htext, nil};
+	int *textc[] = {&text_col_back, &text_col_high, &text_col_bord, &text_col_text,
+	              &text_col_htext, nil};
+
+	for(p = csspec; p && *p; p = (p && *p) ? p + 1: p) {
+		if (strncmp("tag=", p, 4) == 0) {
+			p = set_color_scheme_for_entity(tagc, p + 4);
+		} else if (strncmp("text=", p, 5) == 0) {
+			p = set_color_scheme_for_entity(textc, p + 5);
+		} else {
+			fprint(2, "bad color scheme entity: %s\n", p);
+			return;
+		}
+	}
+}
+
+static char*
+set_color_scheme_for_entity(int **ecols, char *csspec)
+{
+	char *p = csspec;
+	int i = 0;
+
+	while(ecols[i] && *p && (*p != ';')) {
+		/* If a current value isn't empty, parse and save it */
+		if (*p != ':') {
+			*ecols[i] = strtol(p, &p, 0);
+			if (*p && (*p != ':') && (*p != ';'))
+				goto err;
+		}
+		if (*p == ':') {
+			p++;
+			i++;
+		}
+	}
+	if (!ecols[i] || ecols[i+1]) {
+		fprint(2, "color codes quantity is wrong: '%s'\n", csspec);
+		return nil;
+	}
+
+	return p;
+err:
+	fprint(2, "wrong color code in: '%s'\n", csspec);
+	return nil;
+}
 
 void
 killprocs(void)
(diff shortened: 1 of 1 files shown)


lego12239:
I used this for change a color scheme from bright to dark, when eyes are tired.

rsc:
Sorry but no. I asked Rob for this 20+ years ago when acme got colors, and he told me nicely but firmly (in a mail that is sadly lost) that he had worked with a professional on them, that if they were wrong they'd be changed, but that options like that were out of the question. I didn't fully understand then, but I've certainly come around. 

Again, sorry, but no. You've got the source, and it sounds like you've already got a solution that works for you, but that complexity is unnecessary in the main tree and not something we want to maintain.

lego12239:
ok. No problem. But can Rob say exactly monitor model where default colors looks like conceived?

iburago:
@rsc, @robpike:

Keeping the particularities of the implementation in this pull request
aside, it is very reasonable to have *a* mechanism for the user to
adjust the color scheme.

The bitmapped Lucida font originally used for GUI in Plan 9 was also
chosen with the help of professionals (in fact, it was created by ones),
but devdraw still allows to customize the font via fontsrv, anyway—and
for a good reason: there no single typeface or font size that work equally
well in all circumstances.

The way people perceive colors differs for various physiological and
psychological reasons, too. There is no “right” or “wrong” color scheme.
Whenever an argument is made—whether with the help of a professional
or not—that a particular choice of colors is optimal, it can be true to life
only statistically, at best, in the average or quantile sense; it is impossible
to verify that against everybody’s eyes.

The fixed-color-scheme approach implicitly assumes that it is all right
to sacrifice the needs of people whose visual comfort is incompatible
with the chosen colors, in favor of implementation purity. It feels as if
they are written off for nothing but code simplicity.

Software is made for people. A bit of **compassion** toward users never
makes it worse; even at the price of a little extra complexity.

(And the need is obviously there. I have seen several implementations
of the Acme color scheme configurability over the years. A good one
that pops to my mind at the moment is by @mariusae [circa 2017][marius-tweet];
see [his gist with the patch][marius-impl].)

[marius-tweet]: https://twitter.com/marius/status/836824901466091520
[marius-impl]: https://gist.github.com/mariusae/bfb2588e54746e012845adbc062b0113

jxy:
As the "world wide web" goes, everything is heading towards pre-defined fonts/colors.

lego12239:
> As the "world wide web" goes, everything is heading towards pre-defined fonts/colors.

Mmm... What does this mean :-)? Browsers doesn't forbid to choose any fonts/colors nor by users nor by site creators. In any way, if this would be true, why we would be follow this strange logic? For example, people with a poor eyesight need more contrast color scheme. People are different and environment where they exist and computers are different, so some mechanism for tunning is important, imho.

lego12239:
@iburago
thanks for these links! I completely forgot about mouse but2 and but3 selection colors and etc :-D. Fixed here - https://github.com/lego12239/plan9port/tree/custom . Also added a possibility to change a text color for mouse but2 and but3 sel.

rsc:
Again, sorry, but no. You've got the source, and it sounds like you've already got a solution that works for you, but that complexity is unnecessary in the main tree and not something we want to maintain.

iburago:
@rsc:

I accept your position, but do not fully understand it.
Specifically, I cannot help, but wonder:

1. What is so complex about parsing a dozen hex values from a string
(say, an environment variable)?

2. Why is it not justified to have that in the main tree, given that it
resolves issues of eye strain and improves readability for people?

3. Why is that considered complex and unnecessary unlike font
customization, which requires an objectively more complex
implementation? Why not to remove the support for custom typefaces,
then? This view of “unnecessary” complexity does not seem to be
consistent.

3. Has Rob expressed his opinion regarding this recently? [You keep
referring][rsc-hn-2012] to that email of his from late nineties, but I could not find
any indications of him expressing this stance anywhere since then. How
do we know that his opinion has not changed in over these 20+ years?
I think, all the people who have to reimplement this feature to accommodate
the color scheme to their vision abilities, would appreciate hearing his reply
to [the above comment][ib-comment].

[rsc-hn-2012]: https://news.ycombinator.com/item?id=4535977
[ib-comment]: https://github.com/9fans/plan9port/pull/381#issuecomment-604130566

wgrr:
On Fri, 27 Mar 2020 12:54:14 -0700
Igor Burago <notifications@github.com> wrote:

> 2. Why is it not justified to have that in the main tree, given that
> it resolves issues of eye strain and improves readability for people?
> 
First of all, there is any evidence that it's resolves issues with eye
strain?
I've been using acme for a few years now and I haven't experienced yet
any eye strain. Furthermore, if you're having vision problems, you
should look for a doctor, changing your theme just so you feel better,
chances are that it's a placebo, which is always dangerous.
As of "improves readability" evidences shows that it's the
opposite, positive polarity characters makes text easier, faster and
more accurately to read, *especially* under dark ambient illumination,
just to throw the mandatory authority quote, it's from Nielsen which is
famous in software usability topic:

"Use colors with high contrast between the text and the background.
Optimal legibility requires black text on white background (so-called
positive text). White text on a black background (negative text) is
almost as good. Although the contrast ratio is the same as for positive
text, the inverted color scheme throws people off a little and slows
their reading slightly. Legibility suffers much more for color schemes
that make the text any lighter than pure black, especially if the
background is made any darker than pure white"

refs:
https://www.ncbi.nlm.nih.gov/pubmed/28166901
https://journals.sagepub.com/doi/abs/10.1177/0018720813515509?journalCode=hfsa&
Designing web usability: The practice of Simplicity (p 125)

—wagner

iburago:
@wgrr:

No claim was made for changing the default color scheme; it works well
for most people in a lot of circumstances. But in certain circumstances,
for some people, it does not, and in those cases they are left with no
easy way to adjust the color scheme to their needs. Nobody here is
making an argument to take away the default color scheme from those
who have no reason to change it.

As to the research you referred to, I am well aware of it. However,
one cannot deduce from it that the current color scheme is the most
comfortable for everybody all the time. There are fundamental limits
to the kind of conclusions it is possible to reach from a statistical
research, in general. Any claim of optimality obtained from such
study holds, by its very nature, only in the average or quantile sense.
That is, even if, in the extreme case, a particular combination of colors
is found to work best for all 50 participants of a study, it does not
imply that there is no people in the general population for whom that
very combination work as well; it can only provide us with some bounds
on the probability of encountering such people in the general
population, at best (given that model assumptions are justified).

Further, there is more to a color scheme than the effect it has on
a person’s reading speed or text recognition time. There are a number
of environmental, physiological, and psychological factors that affect
how a color is perceived. For instance: many people feel less eye strain
with dark-background color schemes when working in a dimly-lit
environment or at night (like @mariusae whose [patch][marius-tweet] I mentioned
before); people with [floaters][wiki-floater] are often advised against using bright
backgrounds to make floaters less disturbing; some people are working
in front of monitors, which they did not choose, that do not represent
the default colors very nicely (e.g., have an annoying greenish tint
to the default yellow background), etc.

[marius-tweet]: https://twitter.com/marius/status/836824901466091520
[wiki-floater]: https://en.wikipedia.org/wiki/Floater

For all the above reasons, it is not very sensible to insist on a single color
scheme for any interface the users stare at for a prolonged time (like
the one of a text editor). Users need a mechanism for adjustment to
accommodate for the (changing) circumstances they find themselves in.
Even while using ed(1), one can adjust the colors of the text and the
background by changing the settings of their terminal emulator.

And even if one would like to stick to high-contrast positive-text
color schemes only, there is more than one such combination. Funnily
enough, one cannot even follow the recommendation of the black text on
white background that you quoted, because there is no way to run Acme
with white background (instead of the default yellow one) without modifying
the source code.

wgrr:
To be clear Igor, I don't think acme has the most optimal theme and you
should use it, I'm standing against your point that a different color
combination would "help" with clinical conditions, I recall such
arguments being made in past in favor of syntax highlighting. IMO themes
and syntax highlight is a personal aesthetical preference, and it feels
fair the maintainer ask you to maintain your own personal preference,
anybody can modify source code, especially programmers.

—wagner

iburago:
@wgrr:

> I'm standing against your point that a different color combination
would "help" with clinical conditions

It is not my point; I have not made any statements to that effect.

> IMO themes and syntax highlight is a personal aesthetical preference

There is a fundamental difference between the interface color scheme
and syntax highlighting. The latter is a layer of formatting that is
additional by its very nature—it can be present or not. The former,
however, is foundational—if text is displayed in an interface, it cannot
*not* be rendered in a certain font and color on a certain background.
That is why the configurability of color scheme is required for more
than just aesthetic reasons; it pertains to the user’s visual comfort
with the interface in general, and might even affect basic readability
for some, depending on their environment.

> and it feels fair the maintainer ask you to maintain your own
personal preference

I have not made an argument that it is unfair. However, as I explained
in my last comment, it goes beyond aesthetic preference.

> anybody can modify source code, especially programmers.

One can even write one’s own version of acme(1) from scratch, but it
still does not imply that the mainline acme(1) would not benefit from
a bit of compassion to its users in the form of some mechanism for
adjusting interface colors.

Besides, why do you not make the same claim regarding, say, font
customization?

PR #382, opened 2020-03-24
acme: validate address range, when writing to addr

neosimsim:
Report an "address out of order" error instead of crashing, e.g. when
running

	9 echo -n '#3,#2'  | 9p write acme/$winid/addr
	9 echo -n 'dot=addr' | 9p write acme/$winid/ctl

or

	9 echo -n '0/yyy/,0/xxx/' | 9p write acme/$winid/addr
	9 echo -n 'dot=addr'      | 9p write acme/$winid/ctl

with `acme/$winid/body`

	xxx yyy

I'm not so happy validating the address twice but validating in src/cmd/acme/addr.c:/^address
seems to be safer. Unfortunately to get a proper error message, I have to validate
again in src/cmd/acme/xfid.c:/^xfidwrite. Maybe one should replace `evalp` with an error string or error object.


neosimsim:
Is there anything I can do to speedup the review?

PR #389, opened 2020-05-05
Update AIX support

bhuntsman:
This lets most everything compile under AIX using IBM's XL/C.
However, the change to lib9/fmt using stdatomic (9505cd1) breaks it as XL/C does not support C11 atomics.


rsc:
Merged by hand, with some changes. Thanks.

PR #392, opened 2020-05-08
all: AIX updates

bhuntsman:
On AIX, in most cases defining _ALL_SOURCE is the right thing to do.  But occasionally it isn't.
This gets sticky because in some cases _ALL_SOURCE is the only feature test macro that causes some things to be defined that we definitely don't want, as is the case with libdiskfs, acme, etc.  In other cases, such as parts of lib9, it's the only feature test macro that defines some things we need in netinet/in.h.  I wasn't able to find a good solution that fits both use cases.  I'm open to suggestions for alternatives.

### DIFF
--- a/bin/9c
+++ b/bin/9c
@@ -137,7 +137,7 @@ case "$tag" in
 		cflags="$ngflags -g"
 		cflags="$cflags -D__sun__ -D__${s}__"
 		;;
-*AIX*)	usexlc
+*AIX*)		usexlc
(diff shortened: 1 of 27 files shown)


rsc:
We simply _cannot_ put AIX-specific tests in all these programs. We need to find a way to modify u.h and _only_ u.h.

rsc:
For concreteness, what is the problem that happens with acme without this change?

PR #394, opened 2020-05-10
AIX updates

bhuntsman:
On AIX, defining _ALL_SOURCE causes a number of conflicts, but we can work around them here.  Affected by the change are:

libdiskfs
acme
diff
rc


bin/9c
```diff
@@ -137,7 +137,7 @@ case "$tag" in
 		cflags="$ngflags -g"
 		cflags="$cflags -D__sun__ -D__${s}__"
 		;;
-*AIX*)	usexlc
+*AIX*)		usexlc
```
bhuntsman, line 140: It looks odd here, but looks right in vi.

rsc:
The fact that you had to work around the word 'class' makes me think xlc is running in C++ mode. Is it? Is there some way to run in C mode?

bhuntsman:
Sadly, no.  xlc_r is C mode, xlC_r is c++.
The conflict is in /usr/include/math.h.  Here is the problematic section:

```C
#ifndef    __cplusplus
extern     int      class();
#else
     /* _class does nothing but calls class subroutine
      * internally. So prototyping like this will
      * still work and avoid compilation errors
      */
extern     int      _class(double);
#endif
```

Here is a result of attempting to compile src/cmd/diff/diffreg.c without that #define in u.h:

```
$ mk
9c  diffreg.c
diffreg.c:84.5: 1506-343 (S) Redeclaration of class differs from previous declaration on line 1462 of "/usr/include/math.h".
diffreg.c:84.5: 1506-382 (I) The type "int*" of identifier class differs from previous type "int()".
mk: 9c  diffreg.c  : exit status=exit(1)
```

rsc:
I think I have picked up all the changes from this sequence, one way or another. Please sync and see what's left that needs fixing. Thanks for all the info.

PR #397, opened 2020-05-18
Fix devdraw and lib9 so it works again in linux and x11

gdiazlo:
I don't know if the recent changes are related to a bigger redesign, but these changes make devdraw and lib9 to build and work on linux/x11 again.

### DIFF
--- a/src/lib9/open.c
+++ b/src/lib9/open.c
@@ -318,7 +318,7 @@ dirreadmax(int fd, Dir **dp, int max)
 				return -1;
 			break;
 		}
-		if(de->d_name[de->d_namlen] != 0)
+		if(de->d_name[_D_EXACT_NAMLEN(de)] != 0)
 			sysfatal("bad readdir");
 		if(de->d_name[0]=='.' && de->d_name[1]==0)
 			continue;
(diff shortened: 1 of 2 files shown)


src/lib9/open.c
```diff
@@ -318,7 +318,7 @@ dirreadmax(int fd, Dir **dp, int max)
 				return -1;
 			break;
 		}
-		if(de->d_name[de->d_namlen] != 0)
+		if(de->d_name[_D_EXACT_NAMLEN(de)] != 0)
```
bhuntsman, line 321: This doesn't work on AIX.
Alternatively, we could add the following line in u.h for the AIX section:

define _D_EXACT_NAMLEN(d) strlen ((d)->d_name)

gdiazlo: Sorry I could only test on Linux with glibc 2.3, I supposed that was a POSIX thing that could make the code work between the variants using d_namlen and d_reclen. Not sure what's the best option here. May be stick with strlen always?

rsc:
Thank you. Fixed lib9 separately and cherry-picked and submitted your X11 fix.

PR #402, opened 2020-05-21
Higher unicode plane

### LINKED ISSUE #401: fontsrv: unicode beyond 0xffff
I was trying a recent change about supporting runes bigger than 0xffff, but so far, I can't get it to work. I don't know if it is related to the fonts I tried, or may be fontsrv needs to be fixed to support them on x11.

I can see there are differences on x11 vs mac, like src/cmd/fontsrv/x11.c:88
```
		if(charcode > 0xffff)
			break;
```

I tried to naively remove that, but no luck :( 

Any hints in how to add support for it on x11/linux? the alt :, shows pjw face instead of the face with tears of joy.
thanks

jxy:
Display emojis!

fix #401

### DIFF
--- a/include/draw.h
+++ b/include/draw.h
@@ -312,7 +312,7 @@ struct Font
 	short		height;	/* max height of image, interline spacing */
 	short		ascent;	/* top of image to baseline */
 	short		width;	/* widest so far; used in caching only */	
-	short		nsub;	/* number of subfonts */
+	ushort		nsub;	/* number of subfonts */
 	u32int		age;	/* increasing counter; used for LRU */
 	int		maxdepth;	/* maximum depth of all loaded subfonts */
 	int		ncache;	/* size of cache */
--- a/man/man3/cachechars.3
+++ b/man/man3/cachechars.3
@@ -165,7 +165,7 @@ struct Font {
 	short	height;	/* max ht of image;interline space*/
 	short	ascent;	/* top of image to baseline */
 	short	width;	/* widest so far; used in caching */
-	short	nsub;	/* number of subfonts */
+	ushort	nsub;	/* number of subfonts */
 	ulong	age;	/* increasing counter; for LRU */
 	int	ncache;	/* size of cache */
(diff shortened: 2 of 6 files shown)


gdiazlo:
Which font did you use to try?

jxy:
On macOS, any code points not in the font are substituted with glyphs from default fonts.  If you use X11, you need the correct fonts.  You can list the code points by

```console
fontsrv -p FONTNAME/28a
```

view the glyphs by,

```console
fontsrv -p FONTNAME/28a/x1f600.bit | img
```

jxy:
It looks like `DejaVuSans` has a few upper planes you can try.

```console
% fontsrv -p DejaVuSans/28a/|9 grep 'x(.|..)....\.bit'
x10300.bit
x1d300.bit
x1d500.bit
x1d580.bit
x1d780.bit
x1ee00.bit
x1f000.bit
x1f080.bit
x1f300.bit
x1f400.bit
x1f600.bit
```

jxy:
Here is what it looks like from fontsrv (upper: X11, lower: macOS).  The upper one is DejaVuSans rendered by X11 fontconfig.  The lower one is the default rendered by macOS.
Screen Shot 2020-05-21 at 10 39 12 AM

gdiazlo:
Cool, with the DejaVuSans it works! thanks!. Do you know if the OSX behavior can be emulated on linux using an adequate `fonts.conf`?  would that be transparent to fontsrv or should be adapted to support such a thing?

jxy:
I suspect fontsrv has to be modified, or an extra layer in the fontpipe that emulates bind(2) between fontsrv and libdraw.

gdiazlo:
After applying this PR and #373 , dejavu fonts are usable again.

rsc:
Applied this with modifications, as of 5f0fa185.
(Had to make fontsrv not so slow, but without making
SubfontSize larger - it's that small to allow use of very
large fonts without overflowing the x width max in the
subfont file.)

PR #403, opened 2020-05-21
ed: handle Unicode beyond the BMP correctly in list mode.

phonologus:
In list mode, `ed` does not display Unicode codepoints beyond the Basic Multingual Plane correctly. For example, the 😀 emoji (codepoint `U01F600`) should be displayed in list mode as something like `\x01f600`. Instead, we get `\xR600`, which is gibberish.

However, fixing display mode to print every non-ASCII character as an eight-glyph string (`\xhhhhhh`) is hugely distracting, as BMP and/or ASCII codepoints predominate most texts, and list mode becomes cluttered with leading zeros.

One solution is to have `ed` display in list mode only the digits actually needed: two digits for non-printing ASCII; four digits for non-ASCII BMP codepoints, and six digits for codepoints beyond the BMP.

However, if we retain the single `\x` sigil, we create ambiguities: is `\x10f600` a single codepoint, or is it a sequence `\x10f6` followed by the two literal ASCII characters `00`? One solution to this is to make use of three distinct sigils: `\xhh` for non-printing ASCII; `\uhhhh` for non-ASCII BMP codepoints, and `\Uhhhhhh` for codepoints beyond the BMP.

This fix implements exactly this protocol, and updates the `ed(1)` manpage accordingly.

### DIFF
--- a/man/man1/ed.1
+++ b/man/man1/ed.1
@@ -441,10 +441,18 @@ a backspace as
 .LR \eb ,
 backslashes as
 .LR \e\e ,
-and non-printing characters as
+and non-printing ASCII characters as
 a backslash, an
 .LR x ,
-and four hexadecimal digits.
+and two hexadecimal digits.
+non-ASCII characters in the Basic Multilingual Plane
+are printed as a backslash, a small
+.LR u ,
+and four hexadecimal digits; and characters above the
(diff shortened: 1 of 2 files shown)


rsc:
Thanks. I pushed this but changed U<6 digits> to U<8 digits> to match Go syntax.

PR #407, opened 2020-05-30
fontsrv: fix undeclared index variable in function x11.c:load().

phonologus:
Build of `fontsrv` fails on Linux/gcc.

The index variable `i` in the `for`-loop in function`void load(XFont *)` in file `X11.c` is undeclared.

This patch supplies the obvious fix.


phonologus:
Fixed in latest commit. This request now redundant.

PR #413, opened 2020-05-31
Add xft support to rio.

pmikkelsen:
Also change the menu spacing to look more like the one
libdraw makes.

The new menu can now look like this when rio is started with `-font "Cantarell-14"`.
2020-06-01-004635_1920x1080_scrot


nicolagi:
Nice. :) Had to add -I/usr/include/freetype2 to X11H in src/cmd/devdraw/mkwsysrules.sh in Ubuntu 18.04.

pmikkelsen:
Thanks, I have added a second commit with your change :)

markvanatten:
On my system this patch compiles correctly, but an attempt to run rio results in its error message 'can't understand depth 0 screen'. Without the patch, rio runs fine.
gcc 10.1.0
Linux 5.6.15-arch1-1
libxft 2.3.3-2

gdiazlo:
> On my system this patch compiles correctly, but an attempt to run rio results in its error message 'can't understand depth 0 screen'. Without the patch, rio runs fine.
> gcc 10.1.0
> Linux 5.6.15-arch1-1
> libxft 2.3.3-2

Seems there is no default font, so this patch makes -font a required argument to rio. @pmikkelsen would you consider adding a plan9port font as default like before?

pmikkelsen:
I am not really using plan9port anymore so it is unlikely I will look any further into this patch :)

PR #414, opened 2020-06-01
devdraw: send a button release event to the client in rpc_bouncemouse

### LINKED ISSUE #356: 9term keeps showing rio's button 3 menu until another button is pressed
When running 9term under rio and right-clicking, the mouse event is sent to rio instead of 9term via bouncemouse(). If nothing is selected in the menu, I don't think that 9term ever gets a button release event, so every time the mouse moves the menu pops up again. If i press another mouse button right after the rio menu closes, or choose to do something with the window (resize, move) there are no problems.

This happens on fedora 31.

I have tried to dig into the code but without luck, as I couldn't find a way to send a button release event to 9term without removing the focus from rio's menu.

Please let me know if there is something I can do to help.

pmikkelsen:
Before, it sometimes happened that the client (9term) lost focus before
the mouse was released, and would therefore still think it was clicked
when it got focus again later, and thus open the rio menu again.

This fixes #356 for me

### DIFF
--- a/src/cmd/devdraw/x11-screen.c
+++ b/src/cmd/devdraw/x11-screen.c
@@ -1755,7 +1755,7 @@ rpc_bouncemouse(Client *c, Mouse m)
 	XWindow dw;
 
 	xlock();
-	e.type = ButtonPress;
+	e.type = ButtonRelease;
 	e.state = 0;
 	e.button = 0;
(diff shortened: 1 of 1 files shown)


pmikkelsen:
On second try there is now an issue where it is not possible to reshape a 9term window if it has focus. I will try again

PR #419, opened 2020-06-02
buffer overflows in libhtml/lex.c and cmd/htmlroff/roff.c

Rei-sen:
_readx() uses length as its argument and not size, getplaindata() should check if j == length -1

### DIFF
--- a/src/cmd/htmlroff/roff.c
+++ b/src/cmd/htmlroff/roff.c
@@ -257,7 +257,7 @@ copyarg(void)
 	int c;
 	Rune *r;
 
-	if(_readx(buf, sizeof(buf)/sizeof(*buf), ArgMode, 0) < 0)
+	if(_readx(buf, nelem(buf), ArgMode, 0) < 0)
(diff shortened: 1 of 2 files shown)


src/cmd/htmlroff/roff.c
```diff
@@ -257,7 +257,7 @@ copyarg(void)
 	int c;
 	Rune *r;
 
-	if(_readx(buf, sizeof buf, ArgMode, 0) < 0)
+	if(_readx(buf, sizeof(buf)/sizeof(*buf), ArgMode, 0) < 0)
```
fhs, line 260: Use `nelem(buf)`?

Rei-sen: fixed

dancrossnyc:
Fixed in 0144f87dc6c7f2f6becbd55519e433a9b36a466f and 4056d6be4d0fca6fc5e6ccfd24ff4785db9fec15.

PR #425, opened 2020-06-30
Add ttf2subf from freetype-plan9

cjacker:
it's Andrey Mirtchovski's ttf2subf to plan9port with minor changes/fixes.

I am a Chinese user. Although rarely using Chinese when coding, but still need to rendering Chinese characters correctly in editor.

I noticed if ACME open text file with some Chinese characters, the font loading process of fontsrv is very slow and will hang ACME for several seconds to several ten seconds. Maybe all CJK (even non-latin1) users will suffer from this problem.

A proper subf font will solve this issue, make it quick and smooth

### DIFF
--- a/src/cmd/ttf2subf/main.c
+++ b/src/cmd/ttf2subf/main.c
@@ -1,7 +1,9 @@
+#include <u.h>
 #include <ft2build.h>
 #include FT_FREETYPE_H
 #include <freetype/ftglyph.h>
 
+#include <libc.h>
 #include <draw.h>
 #include <memdraw.h>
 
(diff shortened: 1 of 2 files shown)


jxy:
You could just save the font files served by fontsrv.

cjacker:
> You could just save the font files served by fontsrv.
Thanks, I mount the fontsrv fs and copy fonts out, It solved my issue.

$fontsrv
$9 mount `namespace`/font /tmp/font

then enter /tmp/font dir, and copy fonts out as needed.

jxy:
You don't actually need to mount it.  Try
```
fontsrv -p LiHeiPro/26a/x4e00.bit|img
```
or something similar.

Issue #379, opened 2020-03-22
mklatinkbd: handle arbitrary unicode in compose sequences


nelk114:
The compose-key sequences (cf. `keyboard(7)`) accepted by `mklatinkbd` are limited to accepting ascii characters as input; I'd like to be able to use sequences containing other unicode characters in my (highly customised — though even the GB layout (I have a british keyboard) includes unaccepted non-ascii such as `£`) keyboard layout.

`.XCompose` makes this quite trivial but unfortunately leaves me with two quite different sets of compose-key bindings, the plan9port ones both less extensive and less to my liking than the ones I've set up for the rest of my system.

Alternatively, if there were some way to have plan9port use the X input method that'd work too.

rsc:
I thought it already could use the X input method. @jxy am I imagining that?

jxy:
I know we support macOS input methods (no completion or emoji from the Touch Bar, nor dictation).  I've no idea about X input method.

cjacker:
@rsc @jxy 

I worked on XIM support for ACME before and almost finished, but still have some minor issues need to be fix(So no pr now, rough patch attached)

The biggest problem is 'How to get cursor(the text cursor, not mouse cursor) position in devdraw/x11-screen.c and update XIM spot(the im candidate list) to follow the cursor position accordingly."

Such as:
```
spot.x = cursor_pos_x;
spot.y = cursor_pos_y;
XSetICValues(xic, XNPreeditAttributes, spotlist, NULL);
```

It seems there is no api to get cursor x/y?  Would you mind give me some clue?


**Patch:**

```
--- plan9port.orig/src/cmd/devdraw/x11-screen.c
+++ plan9port/src/cmd/devdraw/x11-screen.c
@@ -10,6 +10,7 @@
 #include <mouse.h>
 #include <cursor.h>
 #include <thread.h>
+#include <locale.h>
 #include "x11-memdraw.h"
 #include "devdraw.h"
 
@@ -60,6 +61,12 @@
 	rpc_flush
 };
 
+static XIC xic;
+static XIM xim;
+static XPoint spot;
+static XVaNestedList spotlist;
+
+
 static Xwin*
 newxwin(Client *c)
 {
@@ -161,6 +168,8 @@
 	/*
 	 * Connect to X server.
 	 */
+        setlocale(LC_CTYPE, "");
+        XSetLocaleModifiers("");
 	_x.display = XOpenDisplay(NULL);
 	if(_x.display == nil){
[...]

PR #426, opened 2020-06-30
Add X Input Method support to devdraw

cjacker:
These changes add XIM support to devdraw, now acme/sam(all devdraw based utilities) should work well with iBus/Fcitx im framework.

### DIFF
--- a/src/cmd/devdraw/x11-screen.c
+++ b/src/cmd/devdraw/x11-screen.c
@@ -10,6 +10,7 @@
 #include <mouse.h>
 #include <cursor.h>
 #include <thread.h>
+#include <locale.h>
 #include "x11-memdraw.h"
 #include "devdraw.h"
 
@@ -59,6 +60,12 @@ static ClientImpl x11impl = {
 	rpc_bouncemouse,
 	rpc_flush
 };
+
+static XIC xic;
+static XIM xim;
+static XPoint spot;
+static XVaNestedList spotlist;
+
 
 static Xwin*
 newxwin(Client *c)
@@ -161,6 +168,8 @@ gfx_main(void)
 	/*
 	 * Connect to X server.
 	 */
+        setlocale(LC_CTYPE, "");
+        XSetLocaleModifiers("");
 	_x.display = XOpenDisplay(NULL);
 	if(_x.display == nil){
 		disp = getenv("DISPLAY");
@@ -307,6 +316,8 @@ xloop(void)
 		xlock();
 		while(XPending(_x.display)) {
 			XNextEvent(_x.display, &event);
+			if (XFilterEvent(&event, None))
+				  continue;
 			runxevent(&event);
 		}
 	}
@@ -319,6 +330,9 @@ static void
 runxevent(XEvent *xev)
 {
 	int c;
+	char buf[64]; 
+	Rune rbuf[64];
+	Status status;
 	KeySym k;
 	static Mouse m;
 	XButtonEvent *be;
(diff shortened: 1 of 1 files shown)


fhs:
Just tried it out. Very cool. Thanks!

cjacker:
> Just tried it out. Very cool. Thanks!

Would you mind to try the ''cursor following' patch I posted in #379?

Due to the current implementation of devdraw/x11 and acme, I can not find better way to update the IM spot position, but it should work most of the time.

fhs:
@cjacker The IM spot seems to be near where I'm typing, but almost never exactly where I'm typing. It's very unpredictable. Maybe you can put it where I have the mouse pointer? In acme, you have to move your mouse within the window to type inside it anyway, so it'll guarantee that the IM spot is within the window. Also the mouse will be near any selection menu (e.g. in Anthy).

jxy:
would the IM window have a higher chance of covering what you are typing if it follows mouse?

cjacker:
> @cjacker The IM spot seems to be near where I'm typing, but almost never exactly where I'm typing. It's very unpredictable. Maybe you can put it where I have the mouse pointer? In acme, you have to move your mouse within the window to type inside it anyway, so it'll guarantee that the IM spot is within the window. Also the mouse will be near any selection menu (e.g. in Anthy).

@fhs 

I tested with fcitx and ibus, it can always follow the cursor position correctly. maybe it's due to jp im method? I know inputing jp need another composition before committing to screen.

**By the way, I updated the cursor follow patch posted in that issue today** 

test

Before patch update, the im position will only update when string drawn. At the first time you trigger im, the position is not correct. if CR pressed and empy new line inserted, im position still not updated since there is no string drawn.

After patch update, im pos will update when:

1, after rectangle with width = 3 drawn. when such rectangle drawn, the rectangle should be so-called text cursor, actually, it's a vertical line with some decorations. But such a rectangle will drawn everywhere in acme(for example, every tab bar have one).  

2, after string drawn. this should always correct but not enough as I mentioned above.

If you already test the updated patch and still have troubles about cursor following, please comment out below codes in src/cmd/devdraw/devdraw.c. then the im pos will be not updated in time, but when it follows, it should be correct.

If still no lucky, please revert it. 

the im spot can not set to follow the mouse position, it will get a totally random/unpredictable result and jumps within your screen:-D

```
+                       /* text cursor width is always 3,
+                        * but a lot of place will draw such a rectangle,
+                        * for example, every tag bar will have such a cursor.
+                        * So, here is not enough update im spot, still need to update it
+                        * when string drawn.
+                        */
+                       if(Dx(r) == 3)
+                               client->impl->rpc_setimposition(client, r.max.x, r.max.y);

```

cjacker:
> would the IM window have a higher chance of covering what you are typing if it follows mouse?

Yes, It will, always. we can not do that.

fhs:
@cjacker Sorry, I was testing on [edwood](http://github.com/rjkroege/edwood), which is probably over-drawing stuff.
I've tested with acme, 9term, and sam, and it's working as expected. The only problem is of course when you switch window in acme, the IM spot appears in the previous window you were typing in.

ivoidi:
@cjacker 
I use ibus-libpinyin, it like this (always in the lower left corner) :

webwxgetmsgimg

xelxebar:
Are we waiting on anything in particular to get this merged? XIM would be extremely welcome for CJK users, so I'm curious if any technical blocks exist keeping this PR lingering in limbo for so long.

PR #431, opened 2020-07-22
devdraw, libdraw: fix memory leaks by freeing getns() malloced string

1g0rb0hm:
**getns()** returns a pointer to a malloced string. In some cases that malloced string was not freed causing memory leaks.

### DIFF
--- a/src/cmd/devdraw/srv.c
+++ b/src/cmd/devdraw/srv.c
@@ -88,7 +88,7 @@ threadmain(int argc, char **argv)
 void
 gfx_started(void)
 {
-	char *addr;
+	char *ns, *addr;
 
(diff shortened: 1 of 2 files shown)


rsc:
It's hard to worry too much about tiny startup-time memory leaks (as opposed to ongoing one that leak an arbitrary amount of data), but sure this seems OK.

1g0rb0hm:
> It's hard to worry too much about tiny startup-time memory leaks (as opposed to ongoing one that leak an arbitrary amount of data), but sure this seems OK.
> 

Agree. 

Trying to reduce the number of Coverity issues to make it easier to spot new ones.

Thanks for merging.

Issue #436, opened 2020-07-30
touch fails to create files on OpenBSD; maybe because of calling open with O_CREAT | O_TRUNC


falsifian:
The touch command consistently fails on OpenBSD unless the file already exists.

    angel$ 9 touch f
    touch: f: cannot create: Invalid argument

I tried it with plan9port from OpenBSD's ports, which I think uses commit 329831171dd6ef81c113f101093c7b4947381003.

I ran touch in gdb and it looks like it calls OpenBSD's `open` on line 116 of `src/lib9/open.c` with `umode` equal to 0x600 which I think is `O_CREAT | O_TRUNC` and `perm` equal to `0660`.

The [OpenBSD man page for `open`](https://man.openbsd.org/open.2) says at the end: "The O_TRUNC flag requires that one of O_RDWR or O_WRONLY also be specified, else EINVAL is returned."

On the other hand, open on Linux seems to allow `O_CREAT | O_TRUNC`; maybe that's why it works on Linux?

    angel$ uname -a
    OpenBSD angel.falsifian.org 6.7 GENERIC.MP#5 amd64

falsifian:
I opened #439 which fixes it for me. But I think it may leave the underlying issue open.

How is Plan 9's `create` supposed to behave when the argument is `OREAD` but the file already exists? Is it still supposed to truncate it? If so, to get that working on OpenBSD, I guess it would have to be opened with `O_WRITE` first. Is there a way to downgrade the file descriptor to read-only after that?

PR #439, opened 2020-08-08
touch: fix for OpenBSD.

falsifian:
This fixes https://github.com/9fans/plan9port/issues/436

This doesn't necessarily address the underlying issue: calling p9create with
mode = OREAD should probably be allowed, but currently doesn't work on
OpenBSD.

### DIFF
--- a/src/cmd/touch.c
+++ b/src/cmd/touch.c
@@ -54,7 +54,7 @@ touch(int nocreate, char *name)
 		fprint(2, "touch: %s: cannot wstat: %r\n", name);
 		return 1;
 	}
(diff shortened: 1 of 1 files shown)


dancrossnyc:
Thanks: please retitle the commit to read, "touch: fix for OpenBSD."

falsifian:
Done.

gleydsonsoares:
Yes, #439  fixes the issue with `9 touch' here. but I think, the fix at #440 addresses it consistently for remaining occurrences in plan9port/src/ though.

PR #440, opened 2020-08-09
lib9: create() should use O_TRUNC only on O_RDWR or O_WRONLY

gleydsonsoares:
It addresses a better fix for #439 for fixing likely remaining cases:

openbsd$ grep -rn " create(" * | grep OREAD | grep -v DMDIR 
src/cmd/auth/secstore/secuser.c:22:	fd = create(f, OREAD, perm);
src/cmd/disk/mkext.c:161:	fd = create(name, OREAD, mode);
src/cmd/disk/mkfs.c:452:	fd = create(newfile, OREAD, d->mode);
src/cmd/upas/common/libsys.c:102:		fd = create(s_to_c(l->name), OREAD, DMEXCL|0666);
src/cmd/upas/common/libsys.c:823:	fd = create(file, OREAD, perm);
src/cmd/upas/pop3/pop3.c:729:	fd = create(buf, OREAD, 0666);

### DIFF
--- a/src/lib9/open.c
+++ b/src/lib9/open.c
@@ -95,7 +95,10 @@ p9create(char *path, int mode, ulong perm)
 			goto out;
 		fd = open(path, O_RDONLY);
 	}else{
-		umode = (mode&3)|O_CREAT|O_TRUNC;
+		if (mode == OREAD || mode == OEXEC)
+			umode = (mode&3)|O_CREAT;
+		else
+			umode = (mode&3)|O_CREAT|O_TRUNC;
 		mode &= ~(3|OTRUNC);
 		if(mode&ODIRECT){
 			umode |= O_DIRECT;


falsifian:
What if mode is `OEXEC`?

Also, I don't have a lot of experience with Plan 9, so sorry if this is a silly question, but can you confirm this is really how Plan 9's `create` is supposed to behave? If I read the plan9port `create` man page literally, it seems like the file should still be truncated when mode is `OREAD`, although that seems silly. The following text about `create` isn't qualified with anything about the `omode` argument:

    If the file already exists, it is truncated to 0 length...

Should that be qualified with "and `omode` is `OWRITE` or `ORDWR`"?

gleydsonsoares:
Even if it has enough perm, why should it truncate the file if
it's explicitly setting O_RDONLY to open(2)? It does not make
sense.

If the behavior includes changing a file, you should set the
desired flag to open(2), O_RDWR or O_WRONLY, then if the
file exists and you expect to truncate it, you should set also
O_TRUNC so it would be able to truncate it,

Accordingly to Linux manpage:

"O_TRUNC
              If  the  file already exists and is a regular file and the
              access mode allows writing (i.e., is O_RDWR  or  O_WRONLY)
              it  will  be truncated to length 0.  If the file is a FIFO
              or terminal device file,  the  O_TRUNC  flag  is  ignored.
              Otherwise, the effect of O_TRUNC is unspecified."

From OpenBSD open(2) manpage:

"CAVEATS
     The O_TRUNC flag requires that one of O_RDWR or O_WRONLY also be
     specified, else EINVAL is returned."

You have pointed out in #436 , that when you ran `9 touch' on Linux 
you noticed that it works . Why does it work with Linux even man page
saying the opposite? Maybe some Linux internal dance... OpenBSD 
man page is trusted,I'm sure what you read there it's exactly what you read 
in the code, and outcome behavior.

falsifian:
Agreed that truncating with `OREAD` would be strange.

Thanks for the fix. Would it make sense to revert my change to `touch.c`? Feels safer to pass `OREAD`, and I think that would keep it closer to the Plan 9 source.

dancrossnyc:
OTRUNC | OREAD isn't necessarily strange.

Consider some sort of producer/consumer queue thing implemented as two processes: perhaps both are started independently, but they synchronize on a "queue file" or something similar. The consumer opens with OREAD|OTRUNC to discard the queue's previous contents; the producer opens with OWRITE. Is that a weird way to build such a system? Perhaps, but it's not so weird that it's unimaginable.

What Linux/OpenBSD/POSIX happens to do is not particularly relevant when discussing Plan 9 semantics.

gleydsonsoares:
I understand you, but I disagree. what we are discussing here
is not about plan9 semantics it's about "plan9port" compatibility
with the underlying operating system it depends on and runs over.
Mostly compliant with POSIX standards.

It should be almost at least compatible with the underlying
APIs, otherwise p9p can be broken.

OREAD|O_TRUNC over serious UNIX-like operating system like
OpenBSD will fail with EINVAL, like what was happening with `9 touch'
issue James Cook found hard way.

The (undefined) behavior of OREAD|O_TRUNC varies among
implementations. It is safer among UNIX-like OSes by ORed
flags like OWRITE|O_TRUNC when you need truncating
file to 0.

dancrossnyc:
Aha; this is the compatibility shim.

Meh. I guess this is fine. Combine the two commits, though; there's no reason to separate them.

gleydsonsoares:
ping?

PR #442, opened 2020-08-17
Fix buffer overflow in fspread().

gnoack:
Without this fix, fspread is trusting the server to return as much
data as requested, or less. If a server responds with more data
though, fspread writes beyond the bounds of the buffer to fill, which
is passed in by the caller. It depends on the caller of fspread()
where that buffer is, so there are various possible attack vectors.

In the Plan9 kernel, I found this implemented in devmnt.c, where
overly large responses are truncated to the size requested before
copying, so I assume that this strategy works here too.

This also affects fsread() and fsreadn(), which are based on
fspread().

### DIFF
--- a/src/lib9pclient/read.c
+++ b/src/lib9pclient/read.c
@@ -13,6 +13,7 @@ fspread(CFid *fid, void *buf, long n, vlong offset)
 	Fcall tx, rx;
 	void *freep;
 	uint msize;
+	long nr;
 
 	msize = fid->fs->msize - IOHDRSZ;
 	if(n > msize)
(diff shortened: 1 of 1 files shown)


gnoack:
Friendly ping - silence as response to pull requests can sometimes mean "noone saw it" and sometimes "the review is hard". If it's the latter, I'm around and am checking this occasionally. I'm happy to help if you have more questions.

gnoack:
@rsc another friendly ping - I still believe this fix would be relevant?

dancrossnyc (approved):
The change is fine. Please update the commit message to read, "fspread: fix buffer overflow."

Thanks!

gnoack:
(closed and reopened pull request to trigger CI re-run; this failed with issues in homebrew on macOS)

Issue #434, opened 2020-07-28
devdraw: hang on Linux (acme)


therealfun:
I can't reproduce it easily, but it does happen every day.

```
(gdb) thread apply all bt

Thread 2 (Thread 0x7fb0bdd21700 (LWP 2760)):
#0  0x00007fb0bdf30eec in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/libpthread.so.0
#1  0x00005601c64a0d24 in _procsleep (r=0x5601c6c3b218) at pthread.c:60
#2  0x00005601c64a21bb in procscheduler (p=0x5601c6c3b0f0) at thread.c:441
#3  0x00005601c64a1a71 in _threadswitch () at thread.c:259
#4  0x00005601c64a292b in threadqlock (l=0x5601c64d4420 <xlk>, block=0x1, pc=0x5601c6489775) at thread.c:622
#5  0x00005601c64a6e91 in qlock (l=0x5601c64d4420 <xlk>) at qlock.c:56
#6  0x00005601c6489775 in xlock () at x11-screen.c:137
#7  0x00005601c648d595 in __xputsnarf (data=0x7fb0b80d5616 "/usr/share/doc/kdump-tools/copyright\n/usr/share/man") at x11-screen.c:1557
#8  0x00005601c648da24 in rpc_putsnarf (data=0x7fb0b80d5616 "/usr/share/doc/kdump-tools/copyright\n/usr/share/man") at x11-screen.c:1743
#9  0x00005601c6487479 in runmsg (c=0x5601c6c2b170, m=0x7fb0bdd20c90) at srv.c:288
#10 0x00005601c6486e88 in serveproc (v=0x5601c6c2b170) at srv.c:168
#11 0x00005601c64a1f5a in procmain (p=0x5601c6c3b0f0) at thread.c:384
#12 0x00005601c64a0e3a in startprocfn (v=0x5601c6c2ee70) at pthread.c:99
#13 0x00007fb0bdf2afa3 in start_thread () from /lib/libpthread.so.0
#14 0x00007fb0bde5c1af in clone () from /lib/libc.so.6

Thread 1 (Thread 0x7fb0bdd23180 (LWP 2759)):
#0  0x00007fb0bdf30eec in pthread_cond_wait@@GLIBC_2.3.2 () from /lib/libpthread.so.0
#1  0x00005601c64a0d24 in _procsleep (r=0x5601c6c2a3a8) at pthread.c:60
[...]

PR #443, opened 2020-08-25
reorder X11 and clipboard locks

therealfun:
It should fix https://github.com/9fans/plan9port/issues/434


rsc:
It seems like instead of dropping and reclaiming the xlock, which may invalidate whatever invariants the xlock protects, we should swap the other lock orders in rpc\_getsnarf and \_\_xputsnarf, since those are just grabbing both locks at the same time and it is clearly safe as far as invariants to flip the order. And I think those (along with \_xselect) are the only acquisitions of clip.lk that interact with xlock.

PR #449, opened 2020-10-05
Swap x11 clipboard locks

therealfun:
This should fix https://github.com/9fans/plan9port/issues/434.
I haven't found a way to replace the old branch of the previous PR (https://github.com/9fans/plan9port/pull/443).


therealfun:
ping?

PR #451, opened 2020-10-07
rc: implement 9atoms `sep{command} syntax

oridb:
After a bunch of experience using this
syntax in scripts, I want it available
everywhere.

It  makes a good deal of code far simpler
to write. For example,

        nl='
        '

        for(f in `$nl{find . -type f})
                munge $f

is more explicit about what is split than
setting and unsetting $ifs, and doesn't
cause unexpected changes to $ifs for rc
functions or subshells that are run inside
the `{} section.

### DIFF
--- a/man/man1/rc.1
+++ b/man/man1/rc.1
@@ -228,6 +228,8 @@ The value is a single string containing the components of the named variable
 separated by spaces.  A variable with zero elements yields the empty string.
 .HP
 .BI `{ command }
+.HP
+.BI ` "split " { command }
 .br
 .I rc
 executes the
@@ -240,6 +242,8 @@ If
 .B $ifs
 is not otherwise set, its value is
 .BR "'\ \et\en'" .
+In the second form of the command, split is used instead of
+.BR $ifs .
 .HP
 .BI <{ command }
 .HP
--- a/src/cmd/rc/code.c
+++ b/src/cmd/rc/code.c
@@ -141,10 +141,20 @@ outcode(tree *t, int eflag)
 		emitf(Xconc);
 		break;
 	case '`':
+		emitf(Xmark);
+		if(c0){
+			outcode(c0, 0);
+			emitf(Xglob);
+		}else{
+			emitf(Xmark);
+			emitf(Xword);
+			emits(strdup("ifs"));
+			emitf(Xdol);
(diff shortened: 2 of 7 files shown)


rsc:
Sorry, but plan9port is only aiming for compatibility with the original Plan 9, not the many forks.

oridb:
I wasn't proposing it for compatibility, but for utility.

nicolagi:
I don't mind if this gets merged or not, but, to play devil's advocate, I'd say that the rc parser change that enabled use of unquoted equal signs also wasn't for compatibility with original Plan 9, no?

rsc:
It's true that the parser change for = signs was a divergence from Plan 9 rc.
Those are rare and need careful evaluation justification.
In that case, life on Unix means having tons of command lines that use = signs.

This PR is more a whole new feature than a bug fix, and in general I try very hard not to add those.
The justification in this change seemed to be "9atom did it", which doesn't clear the bar for me.

nicolagi:
Ah, I see that now, and it makes perfect sense. Can't argue with that! Thanks for explaining.

oridb:
> The justification in this change seemed to be "9atom did it", which doesn't clear the bar for me.

That's my mistake for not going into more detail. I hinted at it, in the description, but to spell it out more explicitly: correct ifs handling means that functions need to be paranoid. You need to wrap functions with code to save and restore ifs if there's a chance that they can be called from a `\`{}` block wih $ifs changed:

     fn foo {
         oldifs=$ifs
         ifs=' \t'
         do stuff
         $ifs=$oldifs
    }

     oldifs=$ifs
     ifs=';'
     for(split in `{foo | semicolon-separated}) use $split
     ifs=$oldifs

Very few functions do it -- and it's clunky enough that it seems worth keeping it that way. Because this feature removes one of the biggest reasons for manipulating ifs,  that dance would usually get replaced with:

     fn foo { do stuff }
     for(split in `';'{foo | semicolon-separated}) use $split

In addition, because it makes splitting more painless, it becomes more common to split filenames on '\n' appropriately, which makes it more likely people will write scripts that work with spaces in file names.

Anyways, if features aren't wanted, that's fine -- I just wanted to make sure that I'd actually described my reasoning.

PR #471, opened 2021-01-15
fix setpgrp call on OpenBSD

schultzor:
guessing the other BSDs are similar

### DIFF
--- a/src/libthread/daemonize.c
+++ b/src/libthread/daemonize.c
@@ -106,7 +106,11 @@ _threadsetupdaemonize(void)
 	 * Put it in its own process group so that we don't get a SIGHUP
 	 * when the parent exits.
 	 */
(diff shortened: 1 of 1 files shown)


jxy:
Yes for `__FreeBSD__` please.

jonathangray:
setpgid(0, 0) should work without needing the ifdef

pouyatafti:
As @jonathangray points out setpgid(0,0) should work both on Linux and on BSD (tested on NetBSD).

halfwit:
FreeBSD now compiles correctly with this change

anths:
I've created pull request #497 with the simpler setpgrp->setpgid fix (which I've tested on FreeBSD and macOS), as I'm not sure if there's a way to edit pull requests.

dancrossnyc:
Fixed in #497

PR #474, opened 2021-01-26
9pfuse: support MacFUSE >=4

cptaffe:
[MacFUSE 4](https://github.com/osxfuse/osxfuse/releases/tag/macfuse-4.0.0) removes support for passing device fd to the mount command. Adds support for the receiving the fd over a socket instead, and updates command paths and filesystem name.

### DIFF
--- a/bin/mount
+++ b/bin/mount
@@ -20,11 +20,14 @@ case FreeBSD
 	echo 'don''t know how to mount (no fuse)' >[1=2]
 case Darwin
 	version=`{sw_vers -productVersion|cut -d. -f1,2}
+	major_version = `{echo $version|cut -d. -f1}
 	if(sysctl fuse.version >[2]/dev/null |9 grep -si 'fuse.version' ||
 	   sysctl macfuse.version.number >[2]/dev/null |9 grep -si 'fuse.version' ||
 	   sysctl osxfuse.version.number >[2]/dev/null |9 grep -si 'fuse.version' ||
 	   test -d /System/Library/Extensions/fusefs.kext ||
 	   test -d /Library/Filesystems/osxfuse.fs/Contents/Extensions/$version/osxfuse.kext ||
+	   test -d /Library/Filesystems/macfuse.fs/Contents/Extensions/$version/macfuse.kext ||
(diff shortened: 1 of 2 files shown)


dancrossnyc:
Does the older version also support the FD passing mechanism? I'd like to avoid breaking older systems for folks who haven't upgraded yet.

src/cmd/9pfuse/fuse.c
```diff
@@ -837,6 +840,32 @@ mountfuse(char *mtpt)
 		}
 	}
 
+	/* MacFUSE >=4 dropped support for passing fd */
+	if (strcmp(v, "macfuse") == 0) {
```
cptaffe, line 844: @dancrossnyc MacFUSE >=3.3 supports the new mechanism but with different environment variables. My assumption is that since the VFS name `macfuse` is new, `v` will only be set to `macfuse` on >=4. I branch on that; all other VFS names (and therefore older versions) should continue to the old code below the conditional.

bin/mount
```diff
@@ -20,11 +20,14 @@ case FreeBSD
...
 case Darwin
 	version=`{sw_vers -productVersion|cut -d. -f1,2}
+	major_version = `{echo $version|cut -d. -f1}
 	if(sysctl fuse.version >[2]/dev/null |9 grep -si 'fuse.version' ||
 	   sysctl macfuse.version.number >[2]/dev/null |9 grep -si 'fuse.version' ||
 	   sysctl osxfuse.version.number >[2]/dev/null |9 grep -si 'fuse.version' ||
 	   test -d /System/Library/Extensions/fusefs.kext ||
 	   test -d /Library/Filesystems/osxfuse.fs/Contents/Extensions/$version/osxfuse.kext ||
+	   test -d /Library/Filesystems/macfuse.fs/Contents/Extensions/$version/macfuse.kext ||
+	   test -d /Library/Filesystems/macfuse.fs/Contents/Extensions/$major_version/macfuse.kext ||
```
cptaffe, line 30: MacFUSE has a folder `11` but not `11.1` like it does for 10.x, using the major version enables macOS >=11 to work.

PR #477, opened 2021-01-31
9p: parse lines in rdwr command

droyo:
Here is an excerpt from 9p(1) (emphasis mine)

> repeat until end-of-file on standard input: copy a **line** from the
> file to  standard output,  copy a *line* from standard input to the
> file. Print errors, but don't give up. Rdwr is useful for interacting
> with servers like factotum(4).

Here is a simple shell pipeline that you would think, based on the
description above, would be sufficient to fetch a `proto=pass` key
from factotum:

	% 9p rdwr factotum/rpc <<EOF | awk '/^ok/ && NF > 1'
	start proto=pass role=client service=imap server=imap.example.com user=foo@example.com
	read
	EOF

However, this does not work. The `rdwr` subcommand's main loop assumes
that `read()` will read a single line, no more, no less. From a shell
script it is difficult to control when and how much output is flushed
to a pipe.

This change uses `Brdstr` to read the next line from stdin, up to EOF.
With this, the shell pipeline above works as expected. I decided not to
give the same treatment to the file output because I think it is more
likely that someone has a script out there that expects rdwr to give
all available data per "tick", rather than just the next line.

### DIFF
--- a/src/cmd/9p.c
+++ b/src/cmd/9p.c
@@ -302,8 +302,10 @@ void
 xrdwr(int argc, char **argv)
 {
 	char buf[4096];
+	char *p;
 	int n;
 	CFid *fid;
+	Biobuf *b;
 
 	ARGBEGIN{
 	default:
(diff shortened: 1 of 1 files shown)


dancrossnyc:
This looks good to me; can you squash those two commits together, though?

droyo:
squashed

dancrossnyc:
Committed; thanks.

PR #484, opened 2021-02-15
9term: do not call bouncemouse()

### LINKED ISSUE #356: 9term keeps showing rio's button 3 menu until another button is pressed

nicolagi:
Fixes #356.

Rio allows to hit b3 on non-focused windows to bring up its menu,
which is certainly convenient. 9term makes this happen when b3 is
clicked on its window when it's focused, too, which seems an
exception we could do without and which causes #356. (A workaround,
to make the rio menu go away when this patch is not applied, is to
hover to the 9term's scrollbar and click another button there.)

### DIFF
--- a/src/cmd/9term/9term.c
+++ b/src/cmd/9term/9term.c
@@ -231,8 +231,6 @@ mousethread(void *v)
 			goto Send;
 		}else if(mouse->buttons&2)
 			button2menu(w);
-		else
-			bouncemouse(mouse);
 	}
 }
 


pmikkelsen:
Have you ever used rio on real plan9? I know that the rio in plan9port is nothing like the real thing but I think that being able to open rio's menu by a b3 click in 9term is not something which should be removed. The reason it is kinda strange is that 9term is a separate program in plan9port while it is part of rio on real plan9. I suggest you try it out :)

nicolagi:
> Have you ever used rio on real plan9? I know that the rio in plan9port is nothing like the real thing but I think that being able to open rio's menu by a b3 click in 9term is not something which should be removed. The reason it is kinda strange is that 9term is a separate program in plan9port while it is part of rio on real plan9. I suggest you try it out :)

It actually occurred to me after submitting that this bouncemouse thing may be the behavior of rio in 9, though I've never been aware of it.

Since it's here already, I'll leave it for the maintainers to officially reject (or merge, if divergence from 9 is preferable to the bug). (I'm aware you have a PR with a bug fix in libdraw, eh, but I haven't tried it.)

Issue #31, opened 2015-06-03
acme: Kcmd not being passed through on linux


andrewchambers:
Can't trigger copy or paste via keyboard with any combination of ctrl + alt + windows key.

From the code it looks like windows key is the correct combination.

When I add  `printf("key - %d\n", r);` @ https://github.com/9fans/plan9port/blob/3f1c876e14e7eb45c1e17e1f3f4a28d98ff9b35f/src/cmd/acme/text.c#L669 I get
key - 99 // When pressing c
key - 3   // When pressing Ctrl + c
key - 99 // When pressing Windows key + c
key - 99 // When pressing alt + c  (the keypress is sometimes ignored)

mehlon:
[keyboard.h](https://github.com/9fans/plan9port/blob/master/include/keyboard.h#L42) says it's pretty much mac-only.
[This](https://github.com/mehlon/plan9port/commit/f834285f22db897b1a813148394494facbab3727) changes every occurence of Kcmd to a control sequence so you ctrl+c to copy. ([download as patch](https://github.com/mehlon/plan9port/commit/f834285f22db897b1a813148394494facbab3727.patch))

andrewchambers:
How unfortunate that someone would accept mac only patches. Thanks mehlon.

rsc:
We can't map Control to Kcmd because Kcmd+c (copy) and Ctrl+c (stop process) are very different operations. And Alt was taken too, for compose sequences. 

Now that just about everything has a Windows key too, if someone wants to map the Windows key to Kcmd, that would be fine. I don't know how though.

PR #485, opened 2021-02-19
acme: add x11 mappings

meme:
The macOS bindings are present, but the X11 ones are not. Redo is mapped
to ^Y because the shift modifier is not translated.

This patch works, but I'd prefer if we implemented a modifier system
over keys. For example, we remove `Kctl`, `Kcmd`, `Kshift` and `Kalt`
and instead make these "modifier" keys:

```c
enum {
    Mctl= 0x0100,
    Malt=   0x0200,
    Mshift= 0x0400,
    Msuper= 0x0800,
};
```

This way, the switch can be rewritten to be:

```c
case Mctl|Mshift+'z': /* shift-^Z: redo */
```

Please let me know if this is a reasonable approach to this issue and
something you'd prefer, or if this change is adequate.

### DIFF
--- a/src/cmd/acme/text.c
+++ b/src/cmd/acme/text.c
@@ -761,14 +761,17 @@ texttype(Text *t, Rune r)
 		textshow(t, q0, q0, TRUE);
 		return;
 	case Kcmd+'c':	/* %C: copy */
+	case 0x03:	/* ^C: copy */
 		typecommit(t);
 		cut(t, t, nil, TRUE, FALSE, nil, 0);
 		return;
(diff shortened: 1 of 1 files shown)


fhs:
Previous changes like this one (#365, #31) were rejected, and it was [suggested](https://github.com/9fans/plan9port/issues/31#issuecomment-574440289) that it was acceptable to map the Windows key in X11 to `Kcmd`. So, I made PR #479 to do just that and it has been good enough when I don't have a proper mouse on a laptop.

meme:
I don't run Windows and my Super key is used for other things. I'll maintain my own acme fork.

PR #497, opened 2021-04-22
libthread: use setpgid instead of setpgrp

anths:
This setpgrp breaks the build on FreeBSD (and I think all BSDs). This fix is tested on FreeBSD and macOS.

Pull request #471 already exists for this, but the fix isn't right. The correct fix is noted in the comments, but I'm not sure if there's a way to edit a pull request. This is a clean one to get the change in and the build working again.

### DIFF
--- a/src/libthread/daemonize.c
+++ b/src/libthread/daemonize.c
@@ -106,7 +106,7 @@ _threadsetupdaemonize(void)
 	 * Put it in its own process group so that we don't get a SIGHUP
 	 * when the parent exits.
 	 */
-	setpgrp();
+	setpgid(0, 0);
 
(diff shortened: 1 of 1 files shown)


src/libthread/daemonize.c
```diff
@@ -106,7 +106,7 @@ _threadsetupdaemonize(void)
 	 * Put it in its own process group so that we don't get a SIGHUP
 	 * when the parent exits.
 	 */
-	setpgrp();
+	setpgid(0, 0);
```
nicolagi, line 109: Anthony, forgive me for my ignorance, but I've been wondering (since the intial PR) why `rfork(RFNOTEG)` isn't used instead.

PR #503, opened 2021-05-11
Minor glitch: missing print format placeholder.

lootch:


lootch:
I guess I'd like to learn the ropes with something trivial, like adding a %d placeholder that clearly had been accidentally missed. There are three more tweaks, two of them purely cosmetic, lining up horizontal tabs instead of using spaces, mostly. The last one expands ssh-agent's capabilities beyond the present 2048-bit RSA key limit (I thought I'd mildly future proof the fix for up to 8192 bits, but I didn't have the stomach to try it: my equipment is too slow to generate such a key). I have been using 4096 a lot.

I'm planning to improve the docs for ssh-agent, but I thought I'd wait to see how the smaller tweaks would go down.

Lucio.

0intro:
The change looks fine. Could you replace the commit message by something like "auth/ssh-agent: fix missing print format placeholder"? Thanks.

PR #525, opened 2021-10-08
Add wayland support to devdraw

majiru:
There isn't a great way to move the mouse that I've found, so that is left unimplemented. Everything else seems to work fairly well under sway, and I would believe this to hold true for most other wlroots compositors. I would be interested in reports for those running the wayland variants of gnome of KDE, I have been unable to test under those compositors.

### DIFF
--- a/src/cmd/devdraw/mkfile
+++ b/src/cmd/devdraw/mkfile
@@ -39,6 +39,14 @@ latin1.h: $PLAN9/lib/keyboard $O.mklatinkbd
 $O.macargv: $MACARGV
 	$LD -o $target $prereq
 
+xdgshell=/usr/share/wayland-protocols/stable/xdg-shell/xdg-shell.xml
+
+xdg-shell-protocol.c:
+	wayland-scanner private-code < $xdgshell > xdg-shell-protocol.c
+
+xdg-shell-protocol.h:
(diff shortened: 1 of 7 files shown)


The-King-of-Toasters:
I cannot run devdraw, as it immediately exits with `threads in main proc exited w/o threadmaybackground`. I'm running sway 1.6.1+wlroots 0.14.1-2 on Arch Linux. Note that I can run your wayland patch for devdraw.

fhs:
Seem to be working in `sway` after the following change. A bit disappointed it doesn't handle scaling on high dpi display -- that's the main reason I'd want to use wayland natively.
```diff
diff --git a/src/cmd/devdraw/wl-screen.c b/src/cmd/devdraw/wl-screen.c
index 3710b43c..e5a32ead 100644
--- a/src/cmd/devdraw/wl-screen.c
+++ b/src/cmd/devdraw/wl-screen.c
@@ -149,6 +149,7 @@ gfx_main(void)
 {
        clientruning = 1;
        gfx_started();
+       for(;;) sleep(3600*1000);
 }

 Memimage*
```

hdonnay:
I embarked on the same quest before seeing this PR, and have pushed my work to #550. It's stable enough for my daily use, just acme maximized on a monitor all day. It's even got pointer warping working, although I've only tested that under GNOME.

Issue #519, opened 2021-09-07
Compiler throws a warning for u.h sizeof


camsn0w:
/home/sn0w/plan9port/include/u.h:65:42: warning: expression does not compute the number of elements in this array; element type is ‘struct __jmp_buf_tag’, not ‘long int’ [-Wsizeof-array-div]
   65 | typedef long p9jmp_buf[sizeof(sigjmp_buf)/sizeof(long)];
      |                                          ^
/home/sn0w/plan9port/include/u.h:65:42: note: add parentheses around the second ‘sizeof’ to silence this warning

bhuntsman:
Can you include the compiler error in the issue?

Thank you!

benghancock:
I ran into the same issue, so figured I'd chime in on this thread even though it's been open for a while. First things first though: Thanks very much for `plan9port`! I came to know the joys of using Acme thanks to this project.

I ran the `INSTALL` script on a machine running openSUSE the other day and saw a stream of repeated compiler warnings, the most common of which was this one, excerpted with context:

```
>>> cd /usr/local/plan9/src/cmd/devdraw; mk -k install
9c -I/usr/include x11-load.c
/usr/local/plan9/include/u.h:65:42: warning: expression does not compute the number of elements in this array; element type is ‘struct __jmp_buf_tag’, not ‘long int’ [-Wsizeof-array-div]
   65 | typedef long p9jmp_buf[sizeof(sigjmp_buf)/sizeof(long)];
      |                                          ^
/usr/local/plan9/include/u.h:65:42: note: add parentheses around the second ‘sizeof’ to silence this warning
```

My knowledge of C is very shallow, but following the message in the warning, I tried making this change (which I see is already a PR in #520)

```diff
diff --git a/include/u.h b/include/u.h
index 856e10f4..3d779483 100644
--- a/include/u.h
+++ b/include/u.h
@@ -62,7 +62,7 @@ extern "C" {
 #define _NEEDUINT 1
 #define _NEEDULONG 1

-typedef long p9jmp_buf[sizeof(sigjmp_buf)/sizeof(long)];
+typedef long p9jmp_buf[sizeof(sigjmp_buf)/(sizeof(long))];

 #if defined(__linux__)
 #      include <sys/types.h>
```

Re-executing `./INSTALL` afterward successfully silences this warning, and things still seem to work fine (though I admittedly only tested Acme).

Here's the version of `gcc` I was running:

```
[...]

PR #541, opened 2022-02-04
9c: fix silence two gcc warnings

lufia:
```console
$ gcc --version
gcc (GCC) 11.1.0
```

This PR stops two warnings like as:

```
u.h:65:42: warning: expression does not compute the number of elements in this array; element type is ‘struct __jmp_buf_tag’, not ‘long int’ [-Wsizeof-array-div]
   65 | typedef long p9jmp_buf[sizeof(sigjmp_buf)/sizeof(long)];

pipe.c:12:12: warning: argument 1 of type ‘int[2]’ with mismatched bound [-Warray-parameter=]
   12 | p9pipe(int fd[2])
```

### DIFF
--- a/bin/9c
+++ b/bin/9c
@@ -18,6 +18,8 @@ usegcc()
 		-Wno-stringop-truncation \
 		-Wno-stringop-overflow \
 		-Wno-format-truncation \
+		-Wno-sizeof-array-div \
+		-Wno-array-parameter \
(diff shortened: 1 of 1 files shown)


bhuntsman:
The first one is a second possible solution for issue #519
The alternative solution is pull request #520.

lufia:
@bhuntsman I'd not seen them. Thanks.

PR #542, opened 2022-02-07
rio: pass b2 and b3 to clients

nicolagi:
The existing call to XAllowClients to ReplayPointer is ineffective
because the pointer is not frozen at the time.

To make it effective, we change XGrabTutton to grab the pointer
synchronously. In order not to freeze the window manager though, we
need to always call XAllowEvents after processing a mouse event. We
use ReplayPointer if the button isn't b1.

References:
* https://tronche.com/gui/x/xlib/input/XAllowEvents.html
* https://tronche.com/gui/x/xlib/input/XGrabButton.html
* https://tronche.com/gui/x/xlib/input/XGrabPointer.html


bladedvox:
I think you want to add a call to `top(c)` between the `XMapRaised(dpy, c->parent)` and `active(c)` for b2 and b3 so that alt-tab works properly, but other than that this is wonderful!

nicolagi:
Thanks for the feedback, I tried the change but didn't notice a change in behavior, alt-tab seems to work regardless here (though I never used it). Could you elaborate on the bug? Ah looking at the code, perhaps you mean that not calling `top(c)` causes the order of windows cycled through alt-tab to be incorrect?

Re: the patch, it's been working okay for me, but it's not perfect. Sometimes the X session seems to lock up for a few seconds. This happens to me only once a week or so, so I haven't invested in debugging.

bladedvox:
This is a multipart message in MIME format.

------_=_63230a1c14d30728508519c3_=_
MIME-Version: 1.0
Content-Type: multipart/mixed; boundary="----_=_5345c9331af4203b53816742_=_"

This is a multipart message in MIME format.

------_=_5345c9331af4203b53816742_=_
Content-Type: text/plain; charset=UTF-8
Content-Disposition: inline
Content-Transfer-Encoding: quoted-printable

Yes, eactly-  you have to alt-tab twice to correct the order.

Hmm I'll keep a lookout for lockups, I've been using the patch for 36 hours=
 or so and no ill effects yet!

nicolagi ***@***.***> wrote:
> Thanks for the feedback, I tried the change but didn't notice a change in=
 behavior, alt-tab seems to work regardless here (though I never used it). =
Could you elaborate on the bug? Ah looking at the code, perhaps you mean th=
at not calling `top(c)` causes the order of windows cycled through alt-tab =
to be incorrect?
>=20
> Re: the patch, it's been working okay for me, but it's not perfect. Somet=
imes the X session seems to lock up for a few seconds. This happens to me o=
nly once a week or so, so I haven't invested in debugging.



------_=_5345c9331af4203b53816742_=_--

------_=_63230a1c14d30728508519c3_=_
Content-Disposition: attachment; filename=signature.asc
Content-Type: application/pgp-signature
Content-Transfer-Encoding: 7bit

-----BEGIN PGP SIGNATURE-----

iQJDBAABCAAtFiEEkFh6dA+k/6CXFXU4O3+8IhROY5gFAmIuAdEPHG5pbGl4QG5p
bGZtLmNjAAoJEDt/vCIUTmOY7goP/3JkGJ9T9/Vs/XY/qHBO2tDE40JYVcboJxHg
JWpEQPS7QN5rR4cQ2+8fNpXYhESLP4uTHGyr4QTZKPBPNtTJQxmhzQqPveIlfftP
MygZV+2Ke/uCR6fXi3XG5ReqfMTs92FBY9o1xITUlc3sN0OBxUKsrOacNeiWr78N
VvLYmzGDn3gkoLhpRqMzfHWmGBeNCssBoscCH2qp4r/ovbMAWwulNQvFNiuag0jv
RbkaaR0DVJB+EliKXVDiOHCm/jbmzu8yAB7lgXkY8/Of51lMM0CuZFWphVuHV8Je
RGo3XEdXLTctQByVO0pmEsKzcyR1RUWxB6FrbA4hHDncf5BeWfrFIjQlhXGeRYoV
762+SuatvDbjk3c23lOsciy0uEBn2pzF43buD4ruEQQXEXnqzo+/LLemJkLkt0GI
kyGUqdz/c1gD6P8LHdWgqgXxO6rWHVkscbD7aiRINr4WwY3620GWSzWgijzaMY7H
2qeOhztJnTh2I8Mxl9AozMP8zx555PDSN39APvc9b5AkTkiyiEMtNmoi7EaWDZqM
1kkn0Nh6mdK4ftQTszcu6FoZe8I3W6VGhK/r2g2pQ+rFUt9Ghi3rweolUUY+eMmI
hHyntFcTG/FR0R6rDZM+bmsPy2bWBthtPSSjLolGorYBk9f1kusyIXPs6Bn8TSyz
Cp/NbJGJ
=R5zA
-----END PGP SIGNATURE-----

------_=_63230a1c14d30728508519c3_=_--

nicolagi:
Hey @bladedvox did you experience any lockups? I came back here because it happened once today, and it's the first time since our last correspondence here. :)

PR #552, opened 2022-04-30
Reverse search

bd339:
This PR does two things:

1.  Implement a new Kool command. Kool is the reverse of Look. It searches backwards from point.
2. Clean up the search function: removed unnecessary variables and remove code that AFAICT has no performance or correctness purpose.

I know about :-/ and :? but snarfing the needle to the tag and potentially escaping it is an annoyance. It would be more convenient if there was a mouse chord or key modifier for MB3, bypassing the tag completely. Suggestions or feedback?

### DIFF
--- a/src/cmd/acme/exec.c
+++ b/src/cmd/acme/exec.c
@@ -47,6 +47,7 @@ void	indent(Text*, Text*, Text*, int, int, Rune*, int);
 void	xkill(Text*, Text*, Text*, int, int, Rune*, int);
 void	local(Text*, Text*, Text*, int, int, Rune*, int);
 void	look(Text*, Text*, Text*, int, int, Rune*, int);
+void	kool(Text*, Text*, Text*, int, int, Rune*, int);
 void	newcol(Text*, Text*, Text*, int, int, Rune*, int);
 void	paste(Text*, Text*, Text*, int, int, Rune*, int);
 void	put(Text*, Text*, Text*, int, int, Rune*, int);
@@ -83,6 +84,7 @@ static Rune LKill[] = { 'K', 'i', 'l', 'l', 0 };
 static Rune LLoad[] = { 'L', 'o', 'a', 'd', 0 };
 static Rune LLocal[] = { 'L', 'o', 'c', 'a', 'l', 0 };
 static Rune LLook[] = { 'L', 'o', 'o', 'k', 0 };
+static Rune LKool[] = { 'K', 'o', 'o', 'l', 0 };
 static Rune LNew[] = { 'N', 'e', 'w', 0 };
 static Rune LNewcol[] = { 'N', 'e', 'w', 'c', 'o', 'l', 0 };
 static Rune LPaste[] = { 'P', 'a', 's', 't', 'e', 0 };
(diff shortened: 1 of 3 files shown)


mibk:
> I know about :-/ and :? but snarfing the needle to the tag and potentially escaping it is an annoyance. It would be more convenient if there was a mouse chord or key modifier for MB3, bypassing the tag completely. Suggestions or feedback?

Totally agree. The backwards search is so inconvenient (compared to the forward search), that unless I really need it, I just use forward search and cycle through the whole file. I thought about it many times that it would be nice if the shift key was changing the search direction.

bd339:
Since this patch is all about convenience, I agree it would be ideal if shift+MB3 searched backwards. I haven't looked enough at the code to be sure, but it looks to me like Acme uses two threads for mouse and keyboard handling. Therefore it might not be trivial to synchronise them into one event. Are there any examples of keyboard+mouse interaction in acme I can look at?

jxy:
Isn't it the same as `Edit ?search_term`

bd339:
> Isn't it the same as `Edit ?search_term`

Not really imo. The difference between `Kool` and `Edit ?search_term` is the same as the difference between `Look` and `Edit /search_term`. That means you don't have to retype `search_term`, you can select it and use MB2. You also don't have to escape regex chars. Just today I was searching for ^H in Acme. That query must be rewritten to `Edit ?\^H` to do what I want.

mibk:
> You also don't have to escape regex chars. Just today I was searching for ^H in Acme. That query must be rewritten to `Edit ?\^H` to do what I want.

As a different approach to solving the problem of escaping regexp special chars I've once thought about implementing the following feature from the Plan 9 grep (see `grep(1)`):

> An expression starting with `*` will treat the rest of the expression as literal characters.

I think it's a reasonable, small enough feature that would make the experience with Plan 9 tools more consistent.

bd339:
> > An expression starting with `*` will treat the rest of the expression as literal characters.
> 

Interesting, I didn't know that. Wonder how many other people do :D
Or you can use Look (or Kool with this PR). That's what I did and that's what it's there for AFAIK.

In any case, I feel that belongs in a separate PR if someone implements it: I still don't feel that makes backward search sufficiently convenient compared to forward search. In fact it would be still less convienient than Kool, because you still have to re-type/snarf `search_term` in general (and add the `*` which many people may not know about). Acme is the only editor I know that makes searching in reverse considerably harder.

Maybe I should rename this PR: **Easy** reverse search?

mibk:
> In any case, I feel that belongs in a separate PR if someone implements it

Definitely. It's a separate, albeit related issue.

> Since this patch is all about convenience, I agree it would be ideal if shift+MB3 searched backwards.

The shift+MB3 would be really nice. But your approach, as is, might be a good starting point. Just a couple observations:

- I'd find it more convenient if the name of the reverse search command were something like `Lookb`. `Look` is already present in the window tag, so adding a single letter plus clicking on it is more convenient than typing `Kool` somewhere else.
- If this pull request gets accepted, the Acme man page should probably be updated as well. See `$PLAN9/man/man1/acme.1:437`.

rastagraffix:
> - I'd find it more convenient if the name of the reverse search command were so
> mething like `Lookb`. `Look` is already present in the window tag [...]

Bikeshed!  I think RLook (reverse look) is bit more intuitive.  And
still just a single letter to type.

--lyndon

PR #557, opened 2022-06-04
acme/wind.c: redraw window body with bg color if too small for a single line

### LINKED ISSUE #10: Glitch in acme when enlarging a window in less than font->height
In acme, if you enlarge a window in less than font->height its body is not redrawn correctly, duplicating the old text present at that position. The following images show this behaviour:

Before resizing:
screen shot 2014-12-03 at 23 55 50

After resizing:
screen shot 2014-12-03 at 23 55 03

bladedvox:
closes #10 

Title is self explanitory, fixes the age-old bug.

### DIFF
--- a/src/cmd/acme/wind.c
+++ b/src/cmd/acme/wind.c
@@ -235,7 +235,7 @@ winresize(Window *w, Rectangle r, int safe, int keepextra)
 			r1.min.y = min(y, r.max.y);
 			r1.max.y = r.max.y;
 		}else{
-		  draw(screen, r1, textcols[BACK], nil, ZP);
+			draw(screen, r1, textcols[BACK], nil, ZP);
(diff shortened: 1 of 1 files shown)


dancrossnyc:
The change itself looks fine, but could you please fix the indentation? Looks like a tabs vs. spaces thing crept in.

src/cmd/acme/wind.c
```diff
@@ -235,6 +235,7 @@ winresize(Window *w, Rectangle r, int safe, int keepextra)
 			r1.min.y = min(y, r.max.y);
 			r1.max.y = r.max.y;
 		}else{
+		  draw(screen, r1, textcols[BACK], nil, ZP);
```
dancrossnyc, line 238: Fix indentation, please.

PR #567, opened 2022-08-12
troff: support `\~` as an alias for `\ `

g-branden-robinson:
In groff, Heirloom Doctools troff, and neatroff, the `\~` escape
sequence produces an adjustable (paddable) but unbreakable space.
mandoc, which does not perform adjustment or render to typesetters,
supports the escape sequence as a synonym for `\ `, the same as `\0`.

This extension is of long pedigree: groff has supported it for at least
30 years, Heirloom for 17, mandoc for 13, and neatroff for 6.

Do the same as mandoc to prevent mis-rendering of documents using this
escape sequence.  Heirloom Doctools troff, a descendant of Documenter's
Workbench troff, would be a good place to look for a compatible
implementation of the adjustable semantics for this sequence.

### DIFF
--- a/src/cmd/troff/n1.c
+++ b/src/cmd/troff/n1.c
@@ -512,6 +512,8 @@ g0:
 	case '\n':	/* concealed newline */
 		numtabp[CD].val++;
 		goto g0;
+	case '~':	/* Heirloom/groff/neatroff: unbreakable space */
+		; /* fall through */
 	case ' ':	/* unpaddable space */
 		i = UNPAD;
 		goto gx;


g-branden-robinson:
I tested the patch as follows.

```
$ printf 'foo\\~bar\n' | 9 troff > unpatched.trout
[change n1.c]
$ printf 'foo\\~bar\n' | 9 troff > patched.trout
$ diff -u unpatched.trout patched.trout 
--- unpatched.trout     2022-08-12 18:13:52.023987607 -0500
+++ patched.trout       2022-08-12 18:14:02.743938198 -0500
@@ -18,7 +18,7 @@
 H720
 V120
 cf
-33o50o50~33b50a44rn120 0
+33o50o50 25b50a44rn120 0
 x trailer
 V7920
 x stop
```

PR #578, opened 2022-08-31
winwatch: fix double free bug, simplify error handling, reduce X11 calls

markvanatten:
A double free bug caused winwatch to crash sometimes upon closing a window.
The setjmp in the X error handler was not necessary.
The many XFlush and XSync calls, together with the surrounding changes of X error handler, were not necessary.

### DIFF
--- a/src/cmd/rio/winwatch.c
+++ b/src/cmd/rio/winwatch.c
@@ -24,6 +24,7 @@ struct Win {
 
 XDisplay *dpy;
 XWindow root;
+
 Atom net_active_window;
 Reprog *exclude = nil;
 Win *win;
@@ -36,14 +37,13 @@ int showwmnames;
 Font *font;
 Image *lightblue;
 
-XErrorHandler oldxerrorhandler;
 
-enum {
+enum 
+{
 	PAD = 3,
 	MARGIN = 5
 };
 
-static jmp_buf savebuf;
 
 int
 winwatchxerrorhandler(XDisplay *disp, XErrorEvent *xe)
@@ -53,18 +53,14 @@ winwatchxerrorhandler(XDisplay *disp, XErrorEvent *xe)
 	XGetErrorText(disp, xe->error_code, buf, 100);
 	fprint(2, "winwatch: X error %s, request code %d\n",
 	    buf, xe->request_code);
-	XFlush(disp);
-	XSync(disp, False);
-	XSetErrorHandler(oldxerrorhandler);
-	longjmp(savebuf, 1);
-	return(0);  /* Not reached */
+	return 0;  
 }
 
 void*
 erealloc(void *v, ulong n)
 {
 	v = realloc(v, n);
-	if(v==nil)
+	if(v == nil)
 		sysfatal("out of memory reallocating");
 	return v;
 }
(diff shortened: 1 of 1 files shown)


dancrossnyc:
This change sounds great, but there are a number of style changes that a) obscure the main contentful changes, and b) go against plan9 C style; if those parts of the patch could be reverted and the commit message could be prefixed with, `winwatch:` that'd be great.

markvanatten:
I'll be happy to, but should like to ask your advice on C style. I looked at the Plan 9 source for rio, and at Plan 9's man page on style(6), but may well have missed things.

src/cmd/rio/winwatch.c
```diff
@@ -36,14 +37,13 @@ int showwmnames;
 Font *font;
 Image *lightblue;
 
-XErrorHandler oldxerrorhandler;
 
-enum {
+enum 
+{
```
dancrossnyc, line 42: So this is the sort of formatting change I'm talking about; it obscures the main intent of the change with a lot of extraneous noise. If these parts of the patch could be reverted, that'd be great.

src/cmd/rio/winwatch.c
```diff
@@ -53,18 +53,14 @@ winwatchxerrorhandler(XDisplay *disp, XErrorEvent *xe)
 	XGetErrorText(disp, xe->error_code, buf, 100);
 	fprint(2, "winwatch: X error %s, request code %d\n",
 	    buf, xe->request_code);
-	XFlush(disp);
-	XSync(disp, False);
-	XSetErrorHandler(oldxerrorhandler);
-	longjmp(savebuf, 1);
-	return(0);  /* Not reached */
+	return 0;
```
dancrossnyc, line 56: Same here.

src/cmd/rio/winwatch.c
```diff
@@ -53,18 +53,14 @@ winwatchxerrorhandler(XDisplay *disp, XErrorEvent *xe)
...
-	return(0);  /* Not reached */
+	return 0;  
 }
 
 void*
 erealloc(void *v, ulong n)
 {
 	v = realloc(v, n);
-	if(v==nil)
+	if(v == nil)
```
dancrossnyc, line 63: Same here.

etc....

markvanatten:
This commit modifies just the functionality. Editing for style will be done in a further commit. Please advise on stylistic conventions. E.g. in surrounding '==' with spaces I was following the advice on binary operators in Plan 9's style(6).

dancrossnyc:
> This commit modifies just the functionality. Editing for style will be done in a further commit. Please advise on stylistic conventions. E.g. in surrounding '==' with spaces I was following the advice on binary operators in Plan 9's style(6). 

Thanks.

Re: style changes, as unsatisfying as it is, in general I would simply recommend against them for code of this vintage.

The, perhaps sad, fact is that Plan 9 style had very little mechanical support (the `cb` command exists, but didn't seem terribly widely used) and was only loosely adhered to. I've found that large projects that do not strictly adhere to stylistic rules very quickly find those rules devolving into loose guidelines, and ex post facto changes to being code into style adherence become noisy churn or (worse) inadvertently introduce bugs.

For better or worse, the best strategy for dealing with plan9 code generally is to avoid the temptation to make stylistic changes unless they result in misleading code (for example, extra indentation that would lead one to believe that a line is under the control of some conditional when it is not). The best advice is to mimic the existing code.

Perhaps some day we'll bite the bullet and adopt some tool to automate style enforcement, and then we'll blanket run it across the source base, but I doubt that'll happen any time particularly soon.

PR #579, opened 2022-09-03
A: E for acme to be used with $EDITOR

9gay:
bug: it opens files that have spaces in their names, but for some reason prints an error

### DIFF
--- a/bin/A
+++ b/bin/A
@@ -1,0 +1,19 @@
+#!/usr/local/plan9/bin/rc
+if(~ $#* 0) fileids=`{9p read acme/new/ctl | awk '{ print $1 }'}
+
+for(file){
+	fileid=`{9p read acme/new/ctl | awk '{ print $1 }'}
+	fileids=($fileids $fileid)
+	echo name `{realpath $file} | 9p write acme/$fileid/ctl
+	echo get | 9p write acme/$fileid/ctl
+}
+
+while(sleep 1){
+	currentids=`{9p read acme/index | awk '{ print $1 }'}
+	fileidstmp=()
+	for(fileid in $fileids){
+		if(~ $fileid $currentids) fileidstmp=($fileidstmp $fileid)
+	}
+	fileids=$fileidstmp
+	if(~ $#fileids 0) exit
+}
--- a/man/man1/acme.1
+++ b/man/man1/acme.1
@@ -38,6 +38,11 @@ acme, win, awd \- interactive text windows
 [
 .I label
 ]
+.LP
+.B A
+[
+.I file
+\&... ]
 .SH DESCRIPTION
 .I Acme
 manages windows of text that may be edited interactively or by external programs.
@@ -703,6 +708,18 @@ windows.  An example definition is
 .EX
 	fn cd { builtin cd $1 && awd $sysname }
 .EE
+.PP
+.I A
+is a shell-level command that can be used as
+.B $EDITOR
+in a Unix environment.
+It opens
+.I acme
+windows for each
+.I file
+and then does not exit until all those
+.I acme
+windows are closed.
 .SS "Applications and guide files
 In the directory
 .B /acme


dancrossnyc:
Sounds an awful lot like `B`?

9gay:
the difference between `B` and `E` is that `E` waits for file to be modified (not to mention `E` calls `B`) and is meant to be used as `EDITOR=E`, but `A` is about waiting for window in `acme` to be closed even if the file is not modified and is also meant to be used as `EDITOR=A`, `B` is inadequate for `EDITOR=B` unless you intend to not modify any file when other programs call `$EDITOR`

9gay:
also `B` and `E` are for `sam`, but I use `acme` so I've written `A`

dancrossnyc:
B and E will both work just fine with acme, FWIW.

I'm not sure about this, though; it really feels like a personal tool.

9gay:
@dancrossnyc I don't use `sam`, so if I run `E` in `acme`:
```
% E game.dot
plumb: can't open plumb file: dial unix!/tmp/ns.gay.:0/plumb: connect /tmp/ns.gay.:0/plumb: No such file or directory
editing game.dot
^?26274: signal: interrupt
% 
```

9gay:
> B and E will both work just fine with acme, FWIW.

with `sam` running? yes
without `sam` running? no

> I'm not sure about this, though; it really feels like a personal tool.

just like `B` and `E` are personal to `sam`, `A` is personal to `acme`, in order to run `B` and `E` I have to run `sam`, in order to run `A` I have to run `acme`
here's the source code of `B` (why do I copy-paste code that's already in repo?):
```
#!/bin/sh

files=""
dest=""

line=""
if [ $# -gt 0 ]; then
	case "$1" in
	+[0-9]*)
		line="`echo $1 | sed 's/+/:/'`"
		shift
		;;
	esac
fi

if [ $# -eq 0 ]
then
	echo 'usage: B [+line] file...' 2>&1
	exit 1
fi

if [ "x$DISPLAY" = "x" ]
then
	sam="/tmp/.sam.$USER"
else
	if [ "$DISPLAY" = ":0" ]
	then
		DISPLAY=:0.0
	fi
	sam="/tmp/.sam.$USER.$DISPLAY"
fi

if [ ! -p $sam ]
then
	pwd="`pwd`"
	for i
	do
		i="$i$line"
		i=`cleanname -d "$pwd" "$i"`
		plumb -s B -d edit "$i"
	done
else
	for i
	do
		i="$i$line"
		pwd=`pwd`
		file=`cleanname -d $pwd $i`
		line=`echo $file | sed 's/.*://'`
		file=`echo $file | sed 's/:.*//'`
		echo "B $file" >>$sam
		if [ "x$line" != "x" -a "x$line" != "x$file" ]
		then
			echo "$line" >> $sam
		fi
	done
fi
```
and here's the source code of E:
```
#!/bin/sh

# run B but then wait for the file to change.
# great to set as $EDITOR.
# the notion of a file changing is a little weak.

stat=`ls -l $1`
B "$@"
echo editing "$@"
while sleep 1
do
	nstat=`ls -l $1`
	if [ "x$stat" != "x$nstat" ]
	then
		exit
	fi
done
```

dancrossnyc:
On Tue, Sep 6, 2022 at 1:47 PM 9gay ***@***.***> wrote:

> B and E will both work just fine with acme, FWIW.
>
> with sam running? yes
> without sam running? no
>
That doesn't sound right. I cannot reproduce it locally, either on
plan9ports or plan9. Both work just fine, loading a file into acme, without
sam running.

> I'm not sure about this, though; it really feels like a personal tool.
>
> just like B and E are personal to sam, A is personal to acme, in order to
> run B and E I have to run sam, in order to run A I have to run acme
> here's the source code of B (why do I copy-paste code that's already in
> repo?):
>
I think you mean, "specific to" not "personal to" here, but this seems
mistaken. Sorry.

9gay:
> That doesn't sound right. I cannot reproduce it locally, either on
plan9ports or plan9. Both work just fine, loading a file into acme, without
sam running.

now the puzzling piece of code is `if [ ! -p $sam ]` which runs `plumb -s B -d edit "$i"` no matter is `sam` running or not, this is the first time I run `plumber`, but the interesting thing is that `B` doesn't work without `plumber` anymore even if `sam` is running, but alright, my script is not so useful anymore unless there are people like me who want `E` program to stop only when all `acme` windows it created are closed

9gay:
`B` not working without `plumber` but with `sam` running appears to be bug in alpine linux, both busybox shell and bash always return false on `if [ ! -p $sam ]`

PR #593, opened 2022-12-10
acme: use Edit for <, |, > and to trim spaces during Put

swasey:
Programs reading the event file now notice effects on the body from <, |, and Put.

### DIFF
--- a/src/cmd/acme/exec.c
+++ b/src/cmd/acme/exec.c
@@ -243,6 +243,16 @@ execute(Text *t, uint aq0, uint aq1, int external, Text *argt)
 		free(r);
 		return;
 	}
+	/* Prefer Edit's <, |, > */
+	s = skipbl(r, q1-q0, &n);
+	if(n != 0){
+		c = s[0];
+		if(c=='<' || c=='|' || c=='>'){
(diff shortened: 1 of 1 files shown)


src/cmd/acme/exec.c
```diff
@@ -243,6 +243,16 @@ execute(Text *t, uint aq0, uint aq1, int external, Text *argt)
 		free(r);
 		return;
 	}
+	/* Prefer Edit's <, |, > */
```
swasey, line 246: This has an unintended side-effect.

With the original code, clicking with button 2 on something like `|ind` would set the selection to the output of the command. This version sets the selection to the empty string before that output.

I'll sort it out.

swasey: FWIW, the side effect on dot in acme's `Edit |cmd` does not happen in sam. I'm inclined to regard acme's behavior as wrong and fix it, leaving this interpretation of `|cmd` alone.

(The new `Put` code should likely preserve dot.)

swasey: /cc @rsc in case you have strong opinions.

PR #595, opened 2022-12-28
devdraw: Fix mouse wheel scrolling in 1-tick increments on macOS

zakkor:
Fix an issue with mouse wheel scrolling where float delta values were getting truncated into an integer, resulting in slow single-tick wheel scrolls being ignored


andrewrothman:
Tested with a Logitech ERGO M575 and a Logitech MX Master 3 (with its precise magnetic scroll-wheel) as well as the internal laptop trackpad. Working great for me. Thanks!

PR #612, opened 2023-04-01
Use Github Actions for CI

sevan:
Travis-CI is long gone. Use Github Actions instead.
This a basic config for building on Ubuntu & macos, targetting the latest available image from Github.
I tried to add all supported versions but the job sat there for ~15mins waiting for a runner to pick up the jobs before I aborted.
There is no Coverity integration in this workflow yet but this 3rd party plugin might be with investigating (it requires configuring a secret on the repo setting sections).
https://github.com/marketplace/actions/unofficial-coverity-scan


sevan:
@dancrossnyc is this useful?

dancrossnyc:
It is...though I'm loathe to give up coverity.

PR #618, opened 2023-05-19
Dockerfile to build plan9port in Alpine Linux container

JamesParrott:
Hi all.  
I'm adding this for others to show which packages are definitely needed in the host environment to build plan9port, and maybe even spin up a container for themselves easily.  When building from a docker file, certain errors seem to be suppressed. I hadn't noticed perl was being used at first.  So it is possible I've unwittingly removed something essential.  But rc seems to work on the surface.

A .gitattributes file is also included.  This reduces the pain for Windows users building Docker images from a local plan9port repo (otherwise Git may change the line endings to crlf).

I haven't had much joy with https://pkgs.alpinelinux.org/package/edge/community/x86/plan9port (perhaps I hadn't carried out a final configuration step, but it appears some build directories are hardcoded into it). 

This image did work: https://hub.docker.com/r/plan9d/plan9port but I've no clue what was in the hashes it builds from, so I reverse engineered it by diffing apk list--installed with alpine:build-base, and afterwards stripped out the packages unnecessary, simply to compile plan9port, by trial and error.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text=auto
+* text=false
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text=false
+* text eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,9 @@ FROM alpine:3.18
 RUN apk add \
     build-base \
     expat \
-    git-perl \
-    libssl3 \
-    libretls \
     linux-headers \
-    make
+    make \
+    perl
 
 ENV PLAN9=/usr/local/plan9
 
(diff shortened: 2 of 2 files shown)


sevan:
Seems strange to have a Perl requirement, did you confirm if that was needed?

JamesParrott:
> Seems strange to have a Perl requirement, did you confirm if that was needed?

I entirely agree!  It only added about 35MB to a 350-400MB image so I left it in to be safe.

I confirmed a number of calls to perl are definitely in the installation code (by simply using Ctrl+F).  I believe there was even a  commit to specifically call /usr/bin/perl for BSD, relying on my unreliable memory.

I cannot confirm removing perl breaks anything(*) when plan9port is built via a Dockerfile being built.  So if perl's not redundant, and something else is broken without it I'm unaware of, perhaps someoen could add a PR that causes the build process should fail more noisily in the absence of perl?

(*)  I've only tested running rc and calling ls in it

mcepl:
There is also https://github.com/search?q=plan9port+docker&type=repositories

JamesParrott:
Hi Matěj,
Those are other options for those who want a full Debian base image, instead of an Alpine one.  
Best regards,
James

Issue #587, opened 2022-10-22
replace egrep with grep -E


moralground:
Building plan9port on modern Linux produces a log that's basically unreadable because of so many egrep warnings:

    warning: egrep is obsolescent; using grep -E

It should be a pretty simple replace in the whole source tree.

dancrossnyc:
But what about systems where `egrep` is not `grep -E`? A better fix would be to have an egrep wrapper she'll script that `exec`s `grep -E` on Linux.

moralground:
> A better fix would be to have an egrep wrapper shell script that `exec`s `grep -E` on Linux.

That's what egrep currently is on Linux, but with the obnoxious warning. I think the next version of coreutils will outright refuse to run it.

dancrossnyc:
It is unfortunate that coreutils is breaking compatibility with the rest of the open source ecosystem in this way, as thwre are systems other than Linux (and even Linux-based systems that do not use the GNU utilities, such as Alpine). But the shell script wrapper I was referring to would probably be supplied by plan9ports.

macie:
Isn't the issue fixed by commit 564d45b5a55adeb48f69fdecb2bf4ed1f909850b ?

dancrossnyc:
Yeah, this is done now.

PR #622, opened 2023-07-14
use grep -E instead of egrep

lufia:
I noticed *egrep* always output warning:

> egrep: warning: egrep is obsolescent; using grep -E

I heared this warning may be introduced in GNU Grep v3.8.

### DIFF
--- a/bin/9c
+++ b/bin/9c
@@ -53,7 +53,7 @@ quiet()
 	ignore=$ignore'|expanded from macro'
 
 	grep -v '__p9l_autolib_' $1 |
-	egrep -v "$ignore" |
+	grep -E -v "$ignore" |
 	sed 's/ .first use in this function.$//; s/\"\([^\"][^\"]*\)\", line \([0-9][0-9]*\)/\1:\2/g' |
 	$(which uniq) 1>&2  # avoid built-in uniq on SunOS
 }
--- a/bin/9l
+++ b/bin/9l
@@ -146,7 +146,7 @@ then
 					grep '__p9l_autolib_[a-zA-Z0-9+-]*' |
 					sed 's/.*__p9l_autolib_//; s/:.*//' |
 					sort -u |
-					egrep -v '^(thread|draw)$'
+					grep -E -v '^(thread|draw)$'
 				`
 				a="$a $b"
 			done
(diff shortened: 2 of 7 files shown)


dancrossnyc:
Not all systems use GNU grep, however, and the world as a whole does not seem to be in a rush to get rid of `egrep`. I find the warnings annoying; some uses of `egrep` can probably just be replaced with `grep` (without the `-E`) which would, at least, cut down on the noise.

On the other hand, `grep -E` does seem to be in POSIX (https://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html) so it may be sufficiently portable, and POSIX does make a note about the intention to combine `grep`, `egrep` and `fgrep` into a single command. However, I am not sure that everything that plan9port supports has gotten the memo yet.

lufia:
According to https://9fans.github.io/plan9port/

> Linux (x86, x86-64, PowerPC, and ARM), FreeBSD (x86, x86-64), Mac OS X (x86, x86-64, and Power PC), NetBSD (x86 and PowerPC), OpenBSD (x86 and PowerPC), SunOS (x86-64 and Sparc), Dragonfly BSD (x86-64).

I checked grep supports -E option nowadays on several systems:

* Linux: https://www.gnu.org/software/grep/manual/grep.html
* FreeBSD: https://man.freebsd.org/cgi/man.cgi?grep
* macOS(I couldn't find official man page on the web): https://www.unix.com/man-page/osx/1/grep/
* NetBSD: https://man.netbsd.org/grep.1
* OpenBSD: https://man.openbsd.org/grep.1
* Solaris: https://docs.oracle.com/cd/E23823_01/html/816-5165/grep-1.html
* Dragonfly BSD: https://leaf.dragonflybsd.org/cgi/web-man?command=grep&section=1

The -E option means to allow extended regexp in all implementations of grep(1).

macie:
This PR fixes #587

dancrossnyc:
I believe this has been done.

lufia:
@dancrossnyc Thanks! 

ref https://github.com/9fans/plan9port/commit/564d45b5a55adeb48f69fdecb2bf4ed1f909850b

PR #630, opened 2023-09-04
acme: have Dump save both low and high DPI fontnames

matheuristic:
Instead of only saving a window's currently displayed font's name (`t->reffont->f->name`) to the dump file (the current implementation), have Acme's Dump command save that window's combined low DPI and high DPI font names (`t->reffont->f->lodpi->name` and `t->reffont->f->hidpi->name`) when both are available.

This patch enables Acme dumps to better retain window font state.

In particular, it addresses the following font display quirk when using `Dump` and `Load` in Acme on high-DPI screens (200 DPI or more, like macOS Retina screens) without this patch, shown as an example:

1. Suppose we are using Acme on a high DPI screen, say a Retina MacBook screen.
2. Open a non-empty text file, say `/tmp/abc.txt`, in Acme on a high DPI screen type.
3. Execute `Font /mnt/font/AndaleMono/15a/font`. This font is treated as the low DPI font and Acme automatically doubles its point size to `/mnt/font/AndaleMono/30a/font` as the high DPI font. Since we are using a high DPI screen, the high DPI font is displayed.
4. Execute `Dump` to save window state to a dump file `~/acme.dump`. The saved font for the `/tmp/abc.txt` window in the dump file is `/mnt/font/AndaleMono/30a/font` as that's the displayed font.
5. Close then start a fresh instance of Acme.
6. Execute `Load` to restore window state from `~/acme.dump`. The saved font name for the `/tmp/abc.txt` window, `/mnt/font/AndaleMono/30a/font`, is treated as a low DPI font so it is automatically doubled to in point size (i.e., `/mnt/font/AndaleMono/60a/font` is used) for use as the high DPI font, and since we are on a high DPI screen that 60pt font is displayed.

In the example, the displayed font in the `/tmp/abc.txt` window is clearly too large. That is, font display state was restored incorrectly.
   
With this patch applied, in step 4 of the example, the saved font name for the `/tmp/abc.txt` window in the dump file is now 
`/mnt/font/AndaleMono/15a/font,/mnt/font/AndaleMono/30a/font`. So, on `Load` in step 6, the high DPI font `/mnt/font/AndaleMono/30a/font` is used to display text in the file window. That is, display font state was restored correctly.

### DIFF
--- a/src/cmd/acme/rows.c
+++ b/src/cmd/acme/rows.c
@@ -319,7 +319,7 @@ rowdump(Row *row, char *file)
 	int i, j, fd, m, n, start, dumped;
 	uint q0, q1;
 	Biobuf *b;
-	char *buf, *a, *fontname;
+	char *buf, *a, *fontname, *fontfmt, *fontnamelo, *fontnamehi;
 	Rune *r;
 	Column *c;
 	Window *w, *w1;
@@ -394,9 +394,17 @@ rowdump(Row *row, char *file)
 					if(w1->nopen[QWevent])
 						goto Continue2;
 				}
-			fontname = "";
-			if(t->reffont->f != font)
-				fontname = t->reffont->f->name;
+			fontfmt = "%s";
+			fontnamelo = "";
+			fontnamehi = nil;
+			if(t->reffont->f != font){
+				fontnamelo = t->reffont->f->lodpi->name;
+				if(t->reffont->f->hidpi != nil){
+					fontfmt = "%s,%s";
+					fontnamehi = t->reffont->f->hidpi->name;
+				}
+			}
+			fontname = smprint(fontfmt, fontnamelo, fontnamehi);
 			if(t->file->nname)
 				a = runetobyte(t->file->name, t->file->nname);
 			else
@@ -428,6 +436,7 @@ rowdump(Row *row, char *file)
 					100.0*(w->r.min.y-c->r.min.y)/Dy(c->r),
 					w->body.file->b.nc, fontname);
 			}
+			free(fontname);
 			free(a);
 			winctlprint(w, buf, 0);
 			Bwrite(b, buf, strlen(buf));


src/cmd/acme/rows.c
```diff
@@ -396,7 +396,10 @@ rowdump(Row *row, char *file)
 				}
 			fontname = "";
 			if(t->reffont->f != font)
-				fontname = t->reffont->f->name;
+				if(t->reffont->f->hidpi == nil)
```
dancrossnyc, line 399: Suggested edit:

    const char *fontfmt = "%s";

    /* ... */
    if(t->reffont-> != font && t->reffont->f->hidpi == nil)
        fontfmt = "%s,%s";
    fontname = smprint(fontfmt, t->reffont->f->lodpi->name, t->reffont->f->hidpi->name);

    /* ... */
    free(fontname); // Note unconditional free.

matheuristic: Correct me if I'm wrong, but it looks like the main goal is to try for an unconditional `free(fontname)`.
I can see this makes for a less confusing time understanding the code down the road.

I experimented with a few variations but bumped into two scenarios that causes crashes:

- Calling `smprint` with a mismatched number of format specifiers and additional arguments causes a crash.
- `free(fontname)` when `fontname` points to `""` causes a crash.

I added a new commit with the best solution I can think of to retain the unconditional `free(fontname)`, which is to always print the fontname even when `t->reffont->f == font`. Roughly:

```c
/* ... */
if(t->reffont->f->hidpi == nil){
        fontname = smprint("%s", t->reffont->f->lodpi->name);
}else{
        fontname = smprint("%s,%s", t->reffont->f->lodpi->name, t->reffont->f->hidpi->name);
}

/* ... */
free(fontname);
```

The tradeoff here is that the font name is always dumped for all Acme windows, which mainly means a messier-looking dump file but still functions per expectations.

Let me know if you think this resolution works.

dancrossnyc: > Correct me if I'm wrong, but it looks like the main goal is to try for an unconditional `free(fontname)`. I can see this makes for a less confusing time understanding the code down the road.

Correct.

> I experimented with a few variations but bumped into two scenarios that causes crashes:
> 
> * Calling `smprint` with a mismatched number of format specifiers and additional arguments causes a crash.

This surprises me.  I can see that if you have _fewer_ arguments than the format string consumes, but if you have more, I'd expect the extra to be ignored. I tested this on a couple of my systems, and that's how it works locally:

```
: gaja; cat smprint.c
#include <u.h>
#include <libc.h>

void
main()
{
        char *s = smprint("%s", "hi", "bye");
        print("%s\n", s);
        exits(0);
}
: gaja; 9c smprint.c
: gaja; 9l -o smprint smprint.o
: gaja; ./smprint
hi
: gaja;
```

> * `free(fontname)` when `fontname` points to `""` causes a crash.

That does not surprise me; sorry, I missed that. I you go this route, you would also dynamically allocate `""` for `fontname`.  There are a few ways to do that; see below.

> I added a new commit with the best solution I can think of to retain the unconditional `free(fontname)`, which is to always print the fontname even when `t->reffont->f == font`. Roughly:
> 
> ```c
> /* ... */
> if(t->reffont->f->hidpi == nil)
>         fontname = smprint("%s", t->reffont->f->lodpi->name);
> else
>         fontname = smprint("%s,%s", t->reffont->f->lodpi->name, t->reffont->f->hidpi->name);
> 
> /* ... */
> free(fontname);
> ```
> 
> The tradeoff here is that the font name is always dumped for all Acme windows, which mainly means a messier-looking dump file but still functions per expectations.
> 
> Let me know if you think this works.

I think we can introduce another variable to make this slightly cleaner.

```c
    char *fontname, *fname, *fontfmt;

    fontfmt = "%s";
    fname = "";
    if(t->reffont->f != font){
        fname = t->reffont->lodpi->name;
        if(t->reffont->f->hidpi != nil)
            fontfmt = "%s,%s";
    }
    fontname = smprint(fontfmt, fname, t->reffont->f->hidpi);

    // Later...
    free(fontname);
```

Of course, one could also reuse the `fontname` variable as just a pointer in lieu of a second variable:

```c
    char *fontname, *fontfmt;

    fontfmt = "%s";
    fontname = "";
    if(t->reffont->f != font){
        fontname = t->reffont->lodpi->name;
        if(t->reffont->f->hidpi != nil)
            fontfmt = "%s,%s";
    }
    fontname = smprint(fontfmt, fontname, t->reffont->f->hidpi);

    // Later...
    free(fontname);
```

But I don't see much advantage there.

One could also use a more elaborate set of conditionals:

```c
    if(t->reffont->f == font)
        fontname = strdup("");
    else if(t->reffont->f->hidpi != nil)
        fontname = smprint("%s,%s", t->reffont->lodpi->name, t->reffont->f->hidpi->name);
    else
        fontname = smprint("%s", t->reffont->f->lodpi->name);

    // Sometime later....
    free(fontname);
```

I don't particularly care for this, though; personally I prefer the second formulation. I'd be very surprised (and interested in an issue!) if that crashed.

matheuristic: **On the** `smprint` **crashes**:

> This surprises me. I can see that if you have fewer arguments than the format string consumes, but if you have more, I'd expect the extra to be ignored.

You are correct, I dug into the `smprint` crashes I was getting and found it was a rookie mistake on my part. When `t->reffont->f->hidpi` is `nil` (e.g., using low DPI screens or after executing something like `Font 1*/mnt/font/SomeFont/16a/font` in Acme) I was still effectively calling
```c
smprint("%s", t->reffont->f->lodpi->name, t->reffont->f->hidpi->name)
```
which segfaults due to trying to dereference a null pointer (trying to get `t->reffont->f->hidpi->name` but `t->reffont->f->hidpi` is `nil`).

**On the suggested formulations**:

Your second formulation almost works but I needed to avoid the same "deference a null pointer" pitfall mentioned above. The cleanest way I could think of is to use another `char *fontnamehi` variable like in your first formulation but for the second additional argument to `smprint` in order to handle the scenario where `t->reffont->f->hidpi` is `nil`.

```c
    char *fontname, *fontfmt, *fontnamehi;

    // ...
    fontfmt = "%s";
    fontname = "";
    fontnamehi = "";
    if(t->reffont->f != font){
        fontname = t->reffont->f->lodpi->name;
        if(t->reffont->f->hidpi != nil){
            fontfmt = "%s,%s";
            fontnamehi = t->reffont->f->hidpi->name;
        }
    }
    fontname = smprint(fontfmt, fontname, fontnamehi);

    // Later...
    free(fontname);
```

Other approaches I tried to avoid an additional variable (like using conditional operators or a similar trick using boolean operators) ended up making the code look convoluted and ugly.

If the above code block works, let me know and I can update the PR and squash the commits for a cleaner commit history.

If not, any advice on further modifications or anything I missed is welcome.

dancrossnyc: > **On the** `smprint` **crashes**:
> 
> > This surprises me. I can see that if you have fewer arguments than the format string consumes, but if you have more, I'd expect the extra to be ignored.
> 
> You are correct, I dug into the `smprint` crashes I was getting and found it was a rookie mistake on my part. When `t->reffont->f->hidpi` is `nil` (e.g., using low DPI screens or after executing something like `Font 1*/mnt/font/SomeFont/16a/font` in Acme) I was still effectively calling

Ah, that makes sense! Thanks for digging into it.

> ```c
> smprint("%s", t->reffont->f->lodpi->name, t->reffont->f->hidpi->name)
> ```
> 
> which segfaults due to trying to dereference a null pointer (trying to get `t->reffont->f->hidpi->name` but `t->reffont->f->hidpi` is `nil`).
> 
> **On the suggested formulations**:
> 
> Your second formulation almost works but I needed to avoid the same "deference a null pointer" pitfall mentioned above. The cleanest modification I could think of is to use another `char *fontnamehi` variable like in your first formulation but for the second additional argument to `smprint` in order to handle the scenario where `t->reffont->f->hidpi` is `nil`.

Thanks; that is a great fix.

> ```c
>     char *fontname, *fontfmt, *fontnamehi;
> 
>     // ...
>     fontfmt = "%s";
>     fontname = "";
>     fontnamehi = "";
>     if(t->reffont->f != font){
>         fontname = t->reffont->f->lodpi->name;
>         if(t->reffont->f->hidpi != nil){
>             fontfmt = "%s,%s";
>             fontnamehi = t->reffont->f->hidpi->name;
>         }
>     }
>     fontname = smprint(fontfmt, fontname, fontnamehi);
> 
>     // Later...
>     free(fontname);
> ```
> 
> Other approaches I tried to avoid an additional variable (like using conditional operators or a similar trick using boolean operators) ended up making the code look convoluted and ugly.

Seeing this _in situ_ now, I must apologize for pushing the "one less variable" thing; I think having an extra is actually cleaner. With my apologies, may I suggest this?

```c
    char *fontname, *fontfmt, *fontnamelo, *fontnamehi;

    // ...
    fontfmt = "%s";
    fontnamelo = "";
    fontnamehi = nil;
    if(t->reffont->f != font){
        fontnamelo = t->reffont->f->lodpi->name;
        if(t->reffont->f->hidpi != nil){
            fontfmt = "%s,%s";
            fontnamehi = t->reffont->f->hidpi->name;
        }
    }
    fontname = smprint(fontfmt, fontnamelo, fontnamehi);

    // Later...
    free(fontname);
```

The only change here is introducing `fontnamelo` and not reusing `fontname`.

> (_Update: went ahead and updated plus squashed the commits._)

Thanks for that.

> ~If not,~ any advice on further modifications or anything I missed is welcome.

Again, I apologize for suggesting one thing and then going back on it and suggesting another. Otherwise, I think this is good to go in.

matheuristic: No apologies needed! Thanks much for your help, and I think the end solution is much nicer than what we had at the start.

I've updated the PR with your suggested modifications and re-squashed the commits. Please let me know if there's anything else.

dancrossnyc:
This is great! Thank you so much!

dancrossnyc (approved):

PR #635, opened 2023-12-20
ramfs starts

fgergo:
@dancrossnyc 
with modified erealloc(), ramfs works for me(tm).

Other changes:
Maxsize wasn't limiting ramfs memorty usage but per file usage.
-u added back from 4e. Man page amended.

4e man page for reference:
https://9p.io/magic/man2html/4/ramfs

### DIFF
--- a/man/man4/ramfs.4
+++ b/man/man4/ramfs.4
@@ -40,10 +40,6 @@ and 1 will be the network channel from
 .I ramfs
 to the client machine.
 .PP
-The
-.B –u
-option permits ramfs to consume as much memory per file as needed; without it, ramfs will limit it to some arbitrary amount, currently 1GB (a bit small in the 2020s).
-.PP
 This program is useful mainly as an example of how
 to write a user-level file server.
 It can also be used to provide high-performance temporary files.
--- a/man/man4/ramfs.4
+++ b/man/man4/ramfs.4
@@ -40,6 +40,10 @@ and 1 will be the network channel from
 .I ramfs
 to the client machine.
 .PP
+The
+.B –u
+option permits ramfs to consume as much memory per file as needed; without it, ramfs will limit it to some arbitrary amount, currently 1GB (a bit small in the 2020s).
+.PP
 This program is useful mainly as an example of how
 to write a user-level file server.
 It can also be used to provide high-performance temporary files.
--- a/src/cmd/ramfs.c
+++ b/src/cmd/ramfs.c
@@ -12,8 +12,8 @@
 enum
 {
 	OPERM	= 0x3,		/* mask of all permission types in open mode */
-	Nram	= 4096,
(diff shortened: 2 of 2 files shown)


man/man4/ramfs.4
```diff
@@ -40,6 +40,10 @@ and 1 will be the network channel from
 .I ramfs
 to the client machine.
 .PP
+The
+.B –u
+option permits ramfs to consume as much memory per file as needed; without it, ramfs will limit it to some arbitrary amount, currently 1GB (a bit small in the 2020s).
```
dancrossnyc, line 45: Please wrap this line.

src/cmd/ramfs.c
```diff
@@ -202,22 +207,6 @@ main(int argc, char *argv[])
 	r->mtime = r->atime;
 	r->name = estrdup(".");
 
-	r = &ram[1];
```
dancrossnyc, line 205: Why removing all of this? It seems sort of unrelated to your issue.

src/cmd/ramfs.c
```diff
@@ -875,7 +859,7 @@ void *
 erealloc(void *p, ulong n)
 {
 	p = realloc(p, n);
-	if(!p)
+	if(!p && n != 0)
```
dancrossnyc, line 862: So this is a little bit weird, but `realloc` is being changed in C23 and in general, to avoid future complications, I think it's probably best to follow the new standard behavior (even though plan9 realloc is different etc etc).  I'd probably rewrite this as,

    void *
    erealloc(void *p, ulong n)
    {
        if(n == 0){
            free(p);
            return nil;
        }
        p = realloc(p, n);
        if(!p)
            error("out of memory");
        return p;
    }

dancrossnyc (commented):
Thanks! Some comments below. Also, could you please update the commit description to follow the format of other plan9port commits? Something like,

ramfs: update to work on $X

fgergo:
@dancrossnyc I really appreciate for taking a look and for your comments!
I reverted all changes not relevant to making ramfs work and I meant the new commit message to describe that.
ptal

dancrossnyc (approved):
This looks fine. Thanks!

dancrossnyc:
Yeah, sadly, many C programs in general rely on that behavior; indeed, the 1989 C standard specified that was the behavior. Sadly, the standard change, compiler vendors abused the change, and now here we are where it's UB in C'23. :-( That said, I've talked to the C editor (well, one of the co-editors) about it and I sympathize with the reasons that were outlined to me. They're in sort of an impossibly tight situation.

Anyway, there are a few ways forward. For a wrapper like `erealloc`, I'd be in favor of changing that function to handle 0 explicitly (as opposed to changing the call sites). For other uses of `realloc`, it probably makes the most sense to audit them and see what's doing something that may blow up.

Issue #648, opened 2024-04-03
kcmp in cmd/sort.c conflicts with FreeBSD system header file


VoxSciurorum:
Recent development versions of FreeBSD define a function `kcmp` in `<unistd.h>`

```
     kcmp(pid_t pid1, pid_t pid2, int type, uintptr_t idx1, uintptr_t idx2);
```

This conflicts with `int kcmp(Key *ka, Key *kb)` defined in `src/cmd/sort.c` causing a compilation error.

The `kcmp` system call also exists on Linux but there is no C library wrapper to conflict.  kcmp(2): "Glibc does not provide a wrapper for this system call; call it using syscall(2)."

bakul:
kcmp was added to /usr/include/unicode.h post FreeBSD-14.0.

As kcmp is local to src/cmd/sort.c , renaming it to keycmp would fix this without impacting anything else.

danfe:
I'll rename it to `${PORTNAME}_kcmp()` for the time being to unbreak the FreeBSD port, but admittedly `keycmp()` looks more descriptive.  If it's local to `sort.c`, perhaps it could be also declared as `static`?

bakul:
Declaring it static (and no name change) would work.

PR #655, opened 2024-06-12
cmd/sort: Rename kcmp to Kcmp to fix FreeBSD compile error

alexkapshuk:
FreeBSD 14.1 introduced the kcmp(2) system call, which is declaired in unistd.h.

This symbol colides with the kcmp function which is local to sort.c and results in a compile time error.

Rename kcmp to Kcmp to prevent symbol name collision.

### DIFF
--- a/src/cmd/sort.c
+++ b/src/cmd/sort.c
@@ -108,7 +108,7 @@ void	dokey_gn(Key*, uchar*, uchar*, Field*);
 void	dokey_m(Key*, uchar*, uchar*, Field*);
 void	dokey_r(Key*, uchar*, uchar*, Field*);
 void	done(char*);
-int	kcmp(Key*, Key*);
+int	Kcmp(Key*, Key*);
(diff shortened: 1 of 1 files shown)


danfe:
I don't think that `Kcmp` is significantly better than `kcmp` as a symbol name, both are still four-letter names and thus prone to future collisions.

alexkapshuk:
> I don't think that `Kcmp` is significantly better than `kcmp` as a symbol name, both are still four-letter names and thus prone to future collisions.

I see that you've responded in [](https://github.com/9fans/plan9port/issues/648) with the intent to rename kcmp to keycmp.
Fair enough.
Thanks for your input.
Shall I close the PR, or were you going to do it?

danfe:
> Shall I close the PR, or were you going to do it?

I don't have a particular preference or intention on how this problem is going to be handled, so I'd probably leave it up to developers, they should have enough information by now to make the weighted decision.

PR #663, opened 2024-08-17
Change 9pserve default msg size to 8192 byte payload

vestein463:
Default msg size in 9pserve is 24 bytes too low.


vestein463:
Sorry. Only the change for src/cmd/9pserve.c was intended to be enclosed in this pull request.
I am new in github and did not know how to select this from the other commits.

PR #689, opened 2024-12-11
htmlfmt: read file before alloc

Hojun-Cho:
How about read the file firs?

### DIFF
--- a/src/cmd/htmlfmt/html.c
+++ b/src/cmd/htmlfmt/html.c
@@ -22,6 +22,8 @@ loadhtml(int fd)
 	int n;
 	char buf[4096];
 
+	if((n = read(fd, buf, sizeof buf)) <= 0)
+		return nil;	/* empty file */
 	u = emalloc(sizeof(URLwin));
 	u->infd = fd;
 	u->outfd = 1;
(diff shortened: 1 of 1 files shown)


src/cmd/htmlfmt/html.c
```diff
@@ -22,17 +22,18 @@ loadhtml(int fd)
 	int n;
 	char buf[4096];
 
+	if((n = read(fd, buf, sizeof buf)) <= 0)
```
dancrossnyc, line 25: Has this ever been an issue in practice? `htmlfmt` is not meant to be a long-lived program, and when it exits all of its memory will be implicitly freed. The only memory this saves on is a small amount for `u` and `b` and again a small amount in `parsehtml`.

If you want to go down this path, I think I would just stat the file here, and if it's length is 0, `return nil`.  The loop and so forth would not need to change.

Hojun-Cho: > The only memory this saves on is a small amount for `u` and `b` and again a small amount in `parsehtml`.

I agree with you.

> If you want to go down this path, I think I would just stat the file here

I think it's better to keep it that way. Thank you for your insight.
Close this PR.

PR #693, opened 2025-01-29
devdraw: fix cursor warping under XWayland

e0bt:
If devdraw is run under Wayland via XWayland cursor warping doesn't work.

However it's possible to move the cursor if it's not visible: https://github.com/libsdl-org/SDL/issues/9539
I modified devdraw `rpc_setmouse` function to hide cursor before calling `XWarpPointer()` and show it afterwards.

Hiding and showing cursor is handled by XFixes extenstion which requirers including `X11/extensions/Xfixes.h` header and adding `-lXfixes` to CFLAGS.

### DIFF
--- a/src/cmd/devdraw/mkwsysrules.sh
+++ b/src/cmd/devdraw/mkwsysrules.sh
@@ -49,7 +49,7 @@ echo 'X11='$X11
 echo 'X11H='$X11H
 
 if [ $WSYSTYPE = x11 ]; then
-	echo 'CFLAGS=$CFLAGS '$X11H
+	echo 'CFLAGS=$CFLAGS '$X11H -lXfixes
 	echo 'HFILES=$HFILES $XHFILES'
 	XO=`ls x11-*.c 2>/dev/null | sed 's/\.c$/.o/'`
 	echo 'WSYSOFILES=$WSYSOFILES '$XO
--- a/src/cmd/devdraw/x11-inc.h
+++ b/src/cmd/devdraw/x11-inc.h
@@ -16,6 +16,7 @@
 #include <X11/keysym.h>
 #include <X11/IntrinsicP.h>
 #include <X11/StringDefs.h>
+#include <X11/extensions/Xfixes.h>
 #ifdef SHOWEVENT
 #include <stdio.h>
 #include "../rio/showevent/ShowEvent.c"
--- a/src/cmd/devdraw/x11-screen.c
+++ b/src/cmd/devdraw/x11-screen.c
@@ -1384,8 +1384,10 @@ rpc_setmouse(Client *client, Point p)
 	Xwin *w = (Xwin*)client->view;
 
 	xlock();
+	XFixesHideCursor(_x.display, w->drawable);
 	XWarpPointer(_x.display, None, w->drawable, 0, 0, 0, 0, p.x, p.y);
 	XFlush(_x.display);
+	XFixesShowCursor(_x.display, w->drawable);
 	xunlock();
 }
 


eaburns:
If it's useful, I also have a Wayland driver for devdraw that doesn't require Xwayland: https://github.com/eaburns/plan9port/tree/wayland

lneely:
Reporting that this works brilliantly. Would be delighted to see this merged. Thanks @eaburns!

lneely:
@eaburns I tried your wayland branch, and it works brilliantly. If you haven't submitted a PR, please do -- would be very happy to see it merged.

JLanzuisi:
Hi! This is a really nice, simple, workaround. I am having issues building on Fedora 41 tho:
```
>>> cd /home/j/clones/plan9port/src/cmd/devdraw; mk all
9l -o o.devdraw devdraw.o latin1.o mouseswap.o srv.o winsize.o x11-alloc.o x11-cload.o x11-draw.o x11-fill.o x11-get.o x11-keysym2ucs.o x11-load.o x11-pixelbits.o x11-screen.o x11-unload.o
/bin/ld: x11-screen.o: in function `rpc_setmouse':
/home/j/clones/plan9port/src/cmd/devdraw/x11-screen.c:1387:(.text+0x497): undefined reference to `XFixesHideCursor'
/bin/ld: /home/j/clones/plan9port/src/cmd/devdraw/x11-screen.c:1390:(.text+0x4e0): undefined reference to `XFixesShowCursor'
collect2: error: ld returned 1 exit status
mk: 9l -o o.devdraw ...  : exit status=exit(1)
mk: for i in ...  : exit status=exit(1)
mk: for i in ...  : exit status=exit(1)
```
I've made sure that both `libXfixes-devel` and `libXfixes` packages are installed. Also `ldconfig -p` shows
```
libXfixes.so.3 (libc6,x86-64) => /lib64/libXfixes.so.3
libXfixes.so.3 (libc6) => /lib/libXfixes.so.3
libXfixes.so (libc6,x86-64) => /lib64/libXfixes.so
libXfixes.so (libc6) => /lib/libXfixes.so
```
and the header file is at `/usr/include/X11/extensions/Xfixes.h`.

src/cmd/devdraw/mkwsysrules.sh
```diff
@@ -49,7 +49,7 @@ echo 'X11='$X11
 echo 'X11H='$X11H
 
 if [ $WSYSTYPE = x11 ]; then
-	echo 'CFLAGS=$CFLAGS '$X11H
+	echo 'CFLAGS=$CFLAGS '$X11H -lXfixes
```
JLanzuisi, line 52: Ok, so looking at the output of `mk all` in `$PLAN9/src/cmd/devdraw` I saw that the `-lXfixes` flag appeared before the c files, so I changed this line removing the flag here and adding a `echo 'LDFLAGS=$LDFLAGS -lXfixes` line below. With these changes it built for me, and cursor warping works as expected.

nicolagi:
Thanks @e0bt and @JLanzuisi - applied to my copy, it works great!

src/cmd/devdraw/mkwsysrules.sh
sternenseemann, line 52: Same here, this seems to work correctly:

```suggestion
	echo 'CFLAGS=$CFLAGS '$X11H
	echo 'LDFLAGS=$LDFLAGS -lXfixes'
```

eaburns:
KDE provides a Wayland extension that will draw the window decorations for you. My understanding is that the Gnome project takes the stance that the decorations shouldn't be drawn at the Wayland level in the stack, and they expect you to use GTK+ instead of directly using Wayland.

So on KDE you get the decorations, but unfortunately on Gnome you do not. I'm not sure there is a good way forward for direct Wayland on Gnome for the time being.

PR #696, opened 2025-02-05
Add Wayland driver for devdraw

eaburns:


lneely:
I have been using this fork for awhile now, and it works quite well. Recommend merging. :+1:

tusharhero:
Works for me. But it checks for ~$DISPLAY~ instead of ~$WAYLAND_DISPLAY~,
#+begin_src bash
  DISPLAY=$WAYLAND_DISPLAY 9 acme
#+end_src

Fixes it, but perhaps you can make it look for ~$WAYLAND_DISPLAY~ directly.

JLanzuisi:
I've also been testing this branch, great work! 

I've encountered a bug I can't put my finger on: alt-tabbing to a different window and then alt-tabbing back adds spurious characters. I've attached a quick screencast showing the issue.

https://github.com/user-attachments/assets/ac081aa8-fea5-46de-b89c-ace0e51059e1

This is on wayland on gnome 47.

erentar:
Thank you! I can finally use acme!

lobre:
Hey, I have thought a lot about this Wayland backend implementation, trying also to learn about the Wayland world and get a better understanding of how a client should ideally behave.

I am still not 100% sure about everything, but I think the implementation could be laid out a bit differently for a more robust concurrency model.

I think the main thing we could move toward is a single Wayland thread that owns the `wl_display` and all `wl_*` objects, with RPC threads just pushing commands to it. libwayland isn't apparently happy with concurrent access, and right now we still have `wl_display_dispatch()` running while other threads make `wl_*` calls under a lock. A queue + wake fd approach would keep all Wayland interaction serialized and avoid a whole class of races, without really making the code more complex.

Along the same lines, a poll‑based loop on that thread could be more robust. Using `wl_display_prepare_read/read_events` and polling on the Wayland fd plus a wake fd (and clipboard fds when needed) would avoid doing `roundtrip()` or blocking reads inside callbacks. That makes it easier to reason about the event loop and reduces the chance of stalling the compositor.

If input handling and rendering commits also stay on that one thread, we end up with one place where ordering and state changes happen. It matches the "single graphics thread" idea the code already relies on, and it should make behavior more predictable across different compositors.

It would be great to have real Wayland-wizards recommending on whether this is the way of architecturing this Wayland implementation. Just to be sure it is robust enough and "by the book".

If anyone is interested, I have tried to gather everything I read and my set of ideas for how a client implementation would ideally be laid out in [this Gist](https://gist.github.com/lobre/821756100533c9cdea9dd50e81603378), which I have lazily let the LLM format for easy reading.

fhs:
I've been using this PR for a while and it's working well. The only issue I've seen is that if you switch to a different Linux virtual console and come back to the window manager, you lose keyboard and mouse input. I'm using sway on Fedora.

Issue #699, opened 2025-02-10
Acme crashes on AIX 7.3


bhuntsman:
When running Acme on AIX 7.3, the window briefly flashes on the screen, then disappears, and the following message is printed to the terminal:
```
acme: can't post service: Restart the system call
```

This occurs on AIX 7.3, and possibly 7.2 though I haven't tested it on 7.2.  This was also tested with a working build on AIX 7.1 and copied to a 7.3 system and the same behavior observed.

The error is generated from src/cmd/acme/fsys.c:127

Though it is hard to tell where the "Restart the system call" return value is coming from. 

This appears to be an issue related to timing and possibly threads.  If the program is run under the debugger with a breakpoint set in post9pservice, and the post9pservice call is single-stepped through, the window eventually gets displayed and then works.  Further debugging assistance would be greatly appreciated.

bhuntsman:
This definitely looks like a timing issue on AIX 7.3.  The following patch resolves the issue:
```
diff --git a/src/lib9/post9p.c b/src/lib9/post9p.c
index e894ef9e..93710cbd 100644
--- a/src/lib9/post9p.c
+++ b/src/lib9/post9p.c
@@ -41,6 +41,7 @@ post9pservice(int fd, char *name, char *mtpt)
                }
                close(fd);
                w = waitfor(pid);
+               sleep(10);
                if(w == nil)
                        return -1;
                if(w->msg && w->msg[0]){
```

However, I'm not sure that's the best way to go.  On the other hand it shouldn't hurt.

PR #700, opened 2025-02-17
AIX fixes

bhuntsman:
These commits address several issues on AIX:

* Fix the build on newer versions of AIX
* Clean up the errors displayed by 9c/9l
* Update INSTALL to provide an experience closer to what is offered on other platforms.
* Fix the build of fontsrv on AIX
* Prevent rio from being built on AIX as it is permanently incompatible
* Fix an issue with the post9pservice call which causes some programs to crash on AIX 7.3

Unfortunately, this introduces a requirement to install gawk on AIX, however doing so is fairly straightforward.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -65,6 +65,9 @@ DragonFly|*BSD)
 	echo "CFLAGS='-pthread'" >> $PLAN9/config
 	awk=awk
 	;;
+AIX)
+	awk=gawk
+	;;
 *)
 	awk=awk
 	;;
--- a/INSTALL
+++ b/INSTALL
@@ -118,8 +118,8 @@ fi
 if [ `uname` != Darwin ]; then
 	# Determine whether fontsrv X11 files are available.
 	rm -f a.out
-	cc -o a.out -c -Iinclude -I/usr/include -I/usr/local/include -I/usr/include/freetype2 -I/usr/local/include/freetype2 \
-	    -I/usr/X11R7/include -I/usr/X11R7/include/freetype2 \
+	9c -o a.out -c -Iinclude -I/usr/include -I/usr/local/include -I/usr/include/freetype2 -I/usr/local/include/freetype2 \
+	    -I/usr/X11R7/include -I/usr/X11R7/include/freetype2 -I/opt/freeware/include -I/opt/freeware/include/freetype2 \
 	    -I/usr/X11R6/include -I/usr/X11R6/include/freetype2 src/cmd/fontsrv/x11.c >/dev/null 2>&1
 	if [ -f a.out ]; then
 		echo "	fontsrv dependencies found."
--- a/INSTALL
+++ b/INSTALL
@@ -37,6 +37,7 @@ PATH=/bin:/usr/bin:$PLAN9/bin:$PATH export PATH
 [ -z "$PLAN9_TARGET" ] && PLAN9_TARGET="$PLAN9"
 export PLAN9_TARGET
 
+CCVERSFLAG=''
 case `uname` in
 SunOS)
 	awk=nawk
@@ -67,6 +68,7 @@ DragonFly|*BSD)
 	;;
 AIX)
 	awk=gawk
+	CCVERSFLAG=-qversion
 	;;
 *)
 	awk=awk
@@ -139,7 +141,7 @@ if [ -f LOCAL.config ]; then
 fi
 
 echo "* Compiler version:"
-9c -v 2>&1 | grep -v 'Configured with:' | grep -i version | sed 's/^/	/'
+9c -v $CCVERSFLAG 2>&1 | grep -v 'Configured with:' | grep -i version | sed 's/^/	/'
 
 cd src
 if $dobuild; then
--- a/src/cmd/mkfile
+++ b/src/cmd/mkfile
@@ -5,6 +5,7 @@ TARG=`ls *.[cy] *.lx | egrep -v "\.tab\.c$|^x\." | sed 's/\.[cy]//; s/\.lx//'`
 <$PLAN9/src/mkmany
 
 BUGGERED='CVS|faces|factotum|lp|ip|mailfs|upas|vncv|mnihongo|mpm|index|u9fs|secstore|smugfs|snarfer'
+BUGGERED=${BUGGERED}`[ $(uname) = AIX ] && echo '|rio'`
 DIRS=lex `ls -l |sed -n 's/^d.* //p' |egrep -v "^($BUGGERED)$"|egrep -v '^lex$'` $FONTSRV
 
 <$PLAN9/src/mkdirs
--- a/bin/9c
+++ b/bin/9c
@@ -103,6 +103,7 @@ usexlc()
 		-qsuppress=1506-224 \
 		-qsuppress=1506-1300 \
 		-qsuppress=1506-342 \
+		-qsuppress=1506-311 \
 	"
 	cflags="$cflags -g -qdbxextra -qfullpath"
 	cflags="$cflags $CC9FLAGS"
--- a/bin/9l
+++ b/bin/9l
@@ -335,6 +335,9 @@ quiet()
 	ignore=$ignore'|text-based stub'
 	# macOS linker is incessant about reoccurring -l9, -lsec, -lthread:
 	ignore=$ignore'|ld: warning: ignoring duplicate libraries:'
+	# ignore duplicate symbols on AIX
+	ignore=$ignore'|0711-224'
+	ignore=$ignore'|0711-345'
 
 	sed 's/.*: In function `[^:]*: *//' "$1" |
 	egrep -v "$ignore"
--- a/src/cmd/fontsrv/freetyperules.sh
+++ b/src/cmd/fontsrv/freetyperules.sh
@@ -6,7 +6,11 @@ if [ "x$1" = "xx11" ]; then
 	else
 		i=$2
 	fi
-	echo 'CFLAGS=$CFLAGS '$i'/freetype2' 
-        echo 'LDFLAGS=$LDFLAGS -lfontconfig -lfreetype -lz'
+	if [ "x`uname`" = "xAIX" ]; then
+		i="-I/opt/freeware/include"
+		l="-L/opt/freeware/lib"
+	fi
+	echo 'CFLAGS=$CFLAGS '$i' '$i'/freetype2'
+        echo 'LDFLAGS=$LDFLAGS '$l' -lfontconfig -lfreetype -lz'
 fi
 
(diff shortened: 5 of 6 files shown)


INSTALL
```diff
@@ -115,8 +118,8 @@ fi
 if [ `uname` != Darwin ]; then
 	# Determine whether fontsrv X11 files are available.
 	rm -f a.out
-	cc -o a.out -c -Iinclude -I/usr/include -I/usr/local/include -I/usr/include/freetype2 -I/usr/local/include/freetype2 \
-	    -I/usr/X11R7/include -I/usr/X11R7/include/freetype2 \
+	9c -o a.out -c -Iinclude -I/usr/include -I/usr/local/include -I/usr/include/freetype2 -I/usr/local/include/freetype2 \
```
dancrossnyc, line 121: hmm, I don't know about using `9c` here; perhaps just keep this `cc`?

src/cmd/mkfile
```diff
@@ -5,7 +5,9 @@ TARG=`ls *.[cy] *.lx | egrep -v "\.tab\.c$|^x\." | sed 's/\.[cy]//; s/\.lx//'`
 <$PLAN9/src/mkmany
 
 BUGGERED='CVS|faces|factotum|lp|ip|mailfs|upas|vncv|mnihongo|mpm|index|u9fs|secstore|smugfs|snarfer'
-DIRS=lex `ls -l |sed -n 's/^d.* //p' |egrep -v "^($BUGGERED)$"|egrep -v '^lex$'` $FONTSRV
+UNAME=`uname`
+AIXBUGGERED=`if [ "$UNAME" == "AIX" ]; then echo rio; fi`
```
dancrossnyc, line 9: why not append to `BUGGERED` if on AIX?

INSTALL
bhuntsman: I changed it to 9c because 9c sets up  more plan9port environment, much of which we need on AIX.  With this just using cc, this test fails with the following error messages:

```
 cc -o a.out -c -Iinclude -I/usr/include -I/usr/local/include -I/usr/include/freetype2 -I/usr/local/include/freetype2 -I/usr/X11R7/include -I/usr/X11R7/include/freetype2 -I/opt/freeware/include -I/o>
"include/u.h", line 1.1: 1506-046 (S) Syntax error.
"include/u.h", line 1.4: 1506-166 (S) Definition of function See requires parentheses.
"/usr/include/sys/types.h", line 345.3: 1506-273 (E) Missing type in declaration of sigset_t.
"/usr/include/sys/types.h", line 396.9: 1506-046 (S) Syntax error.
"/usr/include/sys/types.h", line 484.9: 1506-166 (S) Definition of function uint64_t requires parentheses.
"/usr/include/sys/types.h", line 484.18: 1506-276 (S) Syntax error: possible missing '{'?
"/usr/include/unistd.h", line 909.43: 1506-282 (S) The type of the parameters must be specified in a prototype.
"/usr/include/unistd.h", line 915.56: 1506-046 (S) Syntax error.
"/usr/include/unistd.h", line 916.59: 1506-275 (S) Unexpected text off_t encountered.
"/usr/include/unistd.h", line 920.64: 1506-275 (S) Unexpected text size_t encountered.
"/usr/include/unistd.h", line 922.16: 1506-166 (S) Definition of function pid_t requires parentheses.
"/usr/include/unistd.h", line 922.42: 1506-276 (S) Syntax error: possible missing '{'?
"/usr/include/stdlib.h", line 114.3: 1506-273 (E) Missing type in declaration of div_t.
"/usr/include/stdlib.h", line 126.8: 1506-166 (S) Definition of function size_t requires parentheses.
"/usr/include/stdlib.h", line 126.15: 1506-276 (S) Syntax error: possible missing '{'?
"/usr/include/sys/signal.h", line 275.9: 1506-046 (S) Syntax error.
"/usr/include/sys/signal.h", line 309.15: 1506-007 (S) "union sigval" is undefined.
"/usr/include/sys/signal.h", line 347.9: 1506-046 (S) Syntax error.
"/usr/include/sys/vm_types.h", line 34.9: 1506-166 (S) Definition of function ulong_t requires parentheses.
"/usr/include/sys/vm_types.h", line 34.25: 1506-276 (S) Syntax error: possible missing '{'?
"/usr/include/sys/vm_types.h", line 148.9: 1506-045 (S) Undeclared identifier vmhandle_t.
"/usr/include/sys/vm_types.h", line 149.9: 1506-045 (S) Undeclared identifier caddr_t.
"/usr/include/sys/vm_types.h", line 151.3: 1506-273 (E) Missing type in declaration of vmaddr_t.
"/usr/include/sys/vm_types.h", line 164.9: 1506-046 (S) Syntax error.
"/usr/include/sys/vm_types.h", line 258.9: 1506-166 (S) Definition of function vmhandle_t requires parentheses.
"/usr/include/sys/vm_types.h", line 258.25: 1506-276 (S) Syntax error: possible missing '{'?
"/usr/include/sys/m_types.h", line 51.3: 1506-273 (E) Missing type in declaration of label_t.
"/usr/include/sys/m_types.h", line 53.9: 1506-166 (S) Definition of function int32long64_t requires parentheses.
"/usr/include/sys/m_types.h", line 53.25: 1506-276 (S) Syntax error: possible missing '{'?
"/usr/include/sys/mstsave.h", line 96.26: 1506-045 (S) Undeclared identifier __kjmpbuf.
"/usr/include/sys/mstsave.h", line 101.25: 1506-275 (S) Unexpected text __curid encountered.
"/usr/include/sys/mstsave.h", line 101.9: 1506-045 (S) Undeclared identifier pid_t.
"include/libc.h", line 1.57: 1506-209 (S) Character constants must end before the end of a line.
"include/libc.h", line 1.34: 1506-076 (W) Character constant 's /sys/include/libc.h.' has more than 4 characters. No more than rightmost 4 characters are used.
"src/cmd/fontsrv/a.h", line 15.89: 1506-209 (S) Character constants must end before the end of a line.
"src/cmd/fontsrv/a.h", line 15.65: 1506-076 (W) Character constant 's lo rune / SubfontSize' has more than 4 characters. No more than rightmost 4 characters are used.
```

We'd need to add several additional changes here to keep it as cc.  I can investigate that if that is the preference.

src/cmd/mkfile
bhuntsman: That is the effect here, but indirectly.  I suppose I could add AIXBUGGERED to BUGGERED if you would prefer to see it done that way.

src/lib9/post9p.c
```diff
@@ -41,6 +41,7 @@ post9pservice(int fd, char *name, char *mtpt)
 		}
 		close(fd);
 		w = waitfor(pid);
+		sleep(10);
```
dancrossnyc, line 44: I know that you know this, but this is clearly the wrong fix. I recognize that this probably masks the symptoms of the underlying bug enough that things work on AIX with this in place, but I'm very hesitant to commit a bandaid like this into the repo.

INSTALL
```diff
@@ -136,7 +139,7 @@ if [ -f LOCAL.config ]; then
 fi
 
 echo "* Compiler version:"
-9c -v 2>&1 | grep -v 'Configured with:' | grep -i version | sed 's/^/	/'
+9c -v -qversion 2>&1 | grep -v 'Configured with:' | grep -i version | sed 's/^/	/'
```
dancrossnyc, line 142: `-qversion` is presumably specific to xlc; with e.g. clang, the output here makes sense, but isn't what I'd normally expect to see:

```
* Compiler version:
        clang: error: unknown argument '-qversion'; did you mean '--version'?
        OpenBSD clang version 16.0.6
```

(Note the error message that sneaks in.)

Perhaps a way around this is add an extra variable in the large case OS statement above that adds this specifically for AIX:

CCVERSFLAG=''
case `uname` in
# ...
AIX)
    # ...
    CCVERSFLAG=-qversion
    ;;
esac

9c -v $CCVERSFLAG 2>&1 | grep ...

src/cmd/mkfile
dancrossnyc: Yes, I think adding to `BUGGERED` would be better.

INSTALL
dancrossnyc: Oh dear.  Ok, I guess that's fine, then.

bhuntsman: Alternatively, we could add a quick uname test and only use 9c on AIX and leave it as-is for the others...

src/lib9/post9p.c
dancrossnyc: Maybe, maybe not.

Looking at what's going on, it appears that, without the sleep, `waitfor` is returning a bogus, nil, value.  Drilling down into `waitfor`, we see that it delegates to `_wait` and the first thing that does is see if `n` is <= 0 and return `nil` if it is. The other cases where it returns `nil` are if it can't tokenize whatever is in `buf`, and if `malloc` fails.  Let's assume `malloc` isn't failing; then if `tokenize` failed, we should see the "couldn't parse wait message" error somewhere and we don't, so it stands to reason that `n <= 0` is probably true. `n` in this case is the return value from `awaitfor`.

`awaitfor` delegates to `_await`, which in turn calls `wait3` or `wait4`, depending on whether `pid4` is -1. There's a comment there that is curiously incomplete; "On Linux, pid==-1 means anyone; on SunOS it's pid==0." That's nice, but there's no special case for SunOS; perhaps the comment is simply a vestige of something that's changed over time. But it's probably irrelevant here; we know that the pid we get here is the PID of a child process created by `fork()` in `post9pservice`, and the error case there was already dealt with. Therefore, we know that `_await` is calling `wait4` in this instance. The the conclusion here is that this code is failing because `_await` is returning 0 or -1.

How that can that happen?  There are three returns from this function, in three separate cases.  Two of those are `strlen`, which can't return -1, and can only return 0 if passed an empty string. But in this case, I don't see how that can happen: the string in question is the result of a copying a formatted string into a buffer, and we can see statically that that buffer isn't empty.  This suggests that `_await` is returning -1 since due to the check against the return value of `wait3` or `wait4`.  Since we know that we're calling `wait4` here, we can deduce that this function is returning -1 because `wait4` is returning 0 or something negative; I'm not a betting man, but if I were, my would be on -1.

I'll bet the, "Restart the system call" message mentioned in #699 is from actually from `wait4`; some random Google searches suggest that that's the strerror string for "ERESTART" on AIX. I don't have access to AIX (and haven't for decades); does the man page for `wait4` mention anything about `ERESTART`?

Something you may try, instead of single-stepping in the debugger, is seeing if you can put a breakpoint on the `return -1` on line 99 of `await.c` and see if you hit it without the `sleep(10)` and without single stepping.

PR #703, opened 2025-03-05
E as editinacme

rakoo:
In the expected case where the files ends up being edited in acme, do as [editinacme](https://github.com/9fans/go/blob/v0.0.7/acme/editinacme/main.go) does: when the window of the file is closed it means we're done. No need for an extra binary in another repository. The old logic is kept as a backup.

In addition, allow line numbers to be specified in the input (as in `E file:13`) and still work with both logics.

### DIFF
--- a/bin/E
+++ b/bin/E
@@ -2,17 +2,26 @@
 
 # run B but then wait for the file to change.
 # great to set as $EDITOR.
-# the notion of a file changing is a little weak.
+# if editing in acme, watch for the moment the window is closed.
+# otherwise the notion of a file changing is a little weak.
 
 stat=`ls -l $1`
 B "$@"
 echo editing "$@"
-while sleep 1
-do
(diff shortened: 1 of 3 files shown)


rakoo:
For reference lazygit [merged](https://github.com/jesseduffield/lazygit/pull/4360) a change that can use `acme` as a preset, and I want to be able to use `E` as my editor in which case the changes there can be used

musvaage:
Is your sed command portable?

rakoo:
Good call, I'll rework it

rakoo:
requires #708

yrk-lab:
Albeit similar in function, `E` and `editinacme` differ quite a bit in how they operate. Sometimes one can be found usable when the other doesn't work, e.g. as a remote command. I'd advocate to rather keep it as is.

PR #706, opened 2025-03-29
cmd/devdraw: avoid use of X button state field

rogpeppe:
This bug is an intermittent annoyance and does not seem likely to be fixed soon. It means that every so often my acme becomes unusable because it behaves as if a mouse button is permanently set.

Avoid that issue by keeping track of mouse button state directly.

While we're about it, fix another annoying issue: under Linux the FocusOut event reset the kstate field but not the kbuttons field, meaning that the buttons state could inappropriately persist. This happened occasionally by accident, but was easy to reproduce: with the mouse button held down, drag outside the window and click Alt-Tab to switch away from the window. We see a KeyDown event followed by FocusOut but no KeyUp event.

Change-Id: I3708316478c80c41806d7157ae9929c6cd84b662

### DIFF
--- a/src/cmd/devdraw/x11-screen.c
+++ b/src/cmd/devdraw/x11-screen.c
@@ -502,6 +502,7 @@ runxevent(XEvent *xev)
 		 * so clear out the keyboard state when we lose the focus.
 		 */
 		_x.kstate = 0;
+		_x.kbuttons = 0;
 		_x.altdown = 0;
 		gfx_abortcompose(w->client);
 		break;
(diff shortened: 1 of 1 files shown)


dancrossnyc (approved):
I usually prefer that Russ look at acme reviews, but this seems simple enough. Thanks!

rogpeppe:
Ha, I stupidly didn't actually include a link to the bug in question! 
Here it is: https://gitlab.freedesktop.org/xorg/xserver/-/issues/1212

PR #709, opened 2025-04-10
Provide follow option to 9p

rakoo:
This is like `tail -f` but for `9p read`

Requires #708


yrk-lab:
I'd suggest to have a separate `9p tail [-f]` rather than cluttering `9p read`.

PR #713, opened 2025-04-29
tmac/tmac.an: define .MR in a groff compatible way

dmitrivereshchagin:
groff 1.23.0 added .MR to its -man macro package.  The NEWS file states that the inclusion of the macro "was prompted by its introduction to Plan 9 from User Space's troff in August 2020."  From d32deab it seems that the name for Plan 9 from User Space's implementation was suggested by groff maintainer G. Brandon Robinson.

Not sure if the intention was to make these definitions compatible, but it would be nice if they were.

Currently, Plan 9 from User Space's .MR expects its second argument to be parenthesized.  groff's .MR does not.  This results in extra parentheses appearing in manual references when viewing Plan 9 from User Space's manual pages on a system using groff.


dancrossnyc (approved):
Eh.... This seems like a reasonable cleanup, and the parentheses in p9p _are_ superfluous, so...sure.

PR #718, opened 2025-06-20
acme: up/down arrow key navigation

cptaffe:
Remap the up and down arrow keys to navigate to the same position on the next or previous lines in the current window, instead of scrolling up or down. The corresponding position on the next and previous lines are calculated based on the number of characters, where `\t` counts as `tabstop` characters.

When using a monospace font, up and down move in a straight line when navigating lines with varying indentation. In my personal testing this performs well in code, and the user no longer needs to move their hand from the keyboard to the mouse, only to move them back to the keyboard after moving up or down a line.

This feature does not attempt to address visual width with regard to variable width fonts, combining characters, zero-width characters, wrapped lines, etc. I think the cursor could end up in between combining characters. This feature does not attempt to remember position or bounce back when the next line is shorter, so going from line 1 position 40 to line 2 of length 30 to line 3 of length 50 will place your cursor on line 3 position 30. This is most noticeable with blank lines.

It may be wise to place this functionality behind a flag.

### DIFF
--- a/src/cmd/acme/text.c
+++ b/src/cmd/acme/text.c
@@ -563,6 +563,42 @@ textbswidth(Text *t, Rune c)
 	return t->q0-q;
 }
 
+/* Distance to BOL */
+int
+textbacklinewidth(Text *t, uint nline)
+{
+	uint q;
+	Rune r;
+
+	q = t->q0;
+	while(q > 0){
+		r = textreadc(t, q-1);
+		if(r == '\n' && !--nline)
+			break;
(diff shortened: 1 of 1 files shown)


shoce:
i hope this will not get merged as default behavior.

yrk-lab:
Please don't merge: this changes the UX considerably.

hahahahacker2009:
Yeah, this is not the Plan 9 way of doing thing.

PR #728, opened 2025-09-06
B: Print usage message to stderr

iamjpn:
B's usage message redirects its stderr to stdout instead of vice versa. Fix this.


dancrossnyc:
Huh. Oops. Thanks!

PR #730, opened 2025-09-14
Fix mouse cursor stack if new window created after colgrow right window

blogdron:
In the forum user [kaldeon](https://www.linux.org.ru/people/kaldeon/profile) show me a bug https://www.linux.org.ru/forum/talks/17709316?cid=18077919

 * video with bug http://0x0.st/Kb4J.30am.mp4  created by user kaldeon

Im expect code and found problem. If for new window `coladd` no have place in colomn, for previos bottom window do `colgrow` for make more space. Problem, after `colgrow`  called `winmousebut` and  one more times **override** mouse position to growed window **before** save mouse position `savemouse` for call `restoremouse` in future and if we close new window, mouse cursor dont jump back to previos window, and stay on last colomn growed window. Just **no need** every time change mouse pose value, to autogrow window. Moreover in this

https://github.com/9fans/plan9port/blob/bafcdddd31811cb9b13f2a984b291446de04a20b/src/cmd/acme/cols.c#L506-L509

This is what is done, which confirms that call  `winmousebut` inside  `colgrow` is a bug

 * video without bug https://0x0.st/Kc-c.ogv created by me 

-----

Im not user plan9port and acme, and use acme first time, I've double-checked many times to see if this change changes behavior anywhere, but as far as I could examine the code, and as far as I could test it all in the editor, there are no regressions.

-----

P.S. Sorry for my English, I wrote half of it myself and just translated some of the text into English :D

### DIFF
--- a/src/cmd/acme/cols.c
+++ b/src/cmd/acme/cols.c
@@ -469,7 +469,6 @@ colgrow(Column *c, Window *w, int but)
 	free(nl);
 	free(ny);
 	c->safe = TRUE;
-	winmousebut(w);
 }
 
 void


spektrokalter:
@rsc @0intro Hello! Could you please take a moment to review the request? Apologies if I tagged the wrong person.

In short, the bug occurs when the mouse cursor doesn't return after deleting a new +Errors window (in fact, any new window). This happens when some windows adjust their dimensions to make room for the new window.

I've tested the fix locally, and it does resolves the problem without causing any other issues.

Thank you for your time and consideration!

spektrokalter:
I've tested it locally and identified an inconsistency.

While I agree that `winmousebut` shouldn't be triggered every time `colgrow` is called, simply removing it from there breaks some cases where it *should* be called. Specifically:

* If +Errors is obscured, running `echo hello` should not only make the body visible but also move the mouse to that window.
* If a window is obscured and we write anything to its body via /mnt/acme/123/body, the mouse should move to that window.
* Similarly, writing "show" to "ctl" should also trigger the mouse movement.

To address these issues, I've added a call to `winmousebut` in `textshow`, like this:

```
	if(t->w!=nil && t->fr.maxlines==0) {
		colgrow(t->col, t->w, 1);
		winmousebut(t->w);
	}
```

Please, review the change and, if you agree, push it.

spektrokalter:
@blogdron There's another small regression (I hope). Previously, when a user changed a font, the mouse would stay in place, but now it jumps. Here's a quick video demonstrating the issue:

https://github.com/user-attachments/assets/aeb1085d-786b-41b0-b5e9-db6da088690b

It seems like `fontx` shouldn't call `winmousebut`. Previously, it didn't call it due to `but` being `-1`.

blogdron:
@spektrokalter regression by this https://github.com/9fans/plan9port/pull/730#issuecomment-330223964 fixed.

PR #766, opened 2026-04-13
fontsrv: improve rendering of slanted fonts (italics) on macOS

yrk-lab:
Improves renderings of Italic fonts, e.g. the [Go font](https://go.dev/blog/go-fonts) (#765).

* Before:

Screenshot 2026-04-13 at 21 54 34

* After:

Screenshot 2026-04-13 at 22 11 55

For comparison, macOS's own rendering (TextEdit) is shown below:

Screenshot 2026-04-13 at 22 29 55

---
The nature of the fix is to include the actual rendering width into the character size maths, while retaining the original logic for keeping the position advancements.

An inspection of the sub font with [tweak(1)](http://9p.io/magic/man2html?man=tweak&sect=1) shows larger _iwidth_ while retaining _width_ values.

* before:
Image

* after:
Screenshot 2026-04-13 at 22 34 18

The remaining glitches at the end of lines, as well as the clipping of the characters upon selection, are due to assumptions in _frselect_ and the direct uses of _frdrawsel_ of _libframe_ in acme, sam and other clients; the assumptions being that the adjacent characters never overlap. Being much, much trickier to address, those fall outside of the scope of the proposed fix.

PR #767, opened 2026-04-22
fontsrv: improve rendering of slanted fonts (italics) on X11

### LINKED ISSUE #765: fontsrv: slanted fonts (italics) are produced with visual artefacts on macOS
When using Italic version of many fonts, e.g. the [Go font](https://go.dev/blog/go-fonts) the characters look truncated:

Image

Viewing individual characters of the sub font with [tweak(1)](http://9p.io/magic/man2html?man=tweak&sect=1) reveals that the images of certain characters are "clipped":

Image

The fix is underway.

yrk-lab:
Improves renderings of Italic fonts, e.g. _URWBookman-DemiItalic_.

Improve renderings of Italic fonts, e.g. the [Go font](https://go.dev/blog/go-fonts).

Before and after, as presented with [tweak(1)](http://9p.io/magic/man2html?man=tweak&sect=1):

20260422_14h26m58s_grim

As before, the nature of the fix is to include the actual rendering width into the character size math, while retaining the original logic of tracking the position advancements: note the larger _iwidth_ values, with the _width_ and _left_ values used for the tracking.

This does not yet address the remaining visual artefacts produced by the programs which make certain assumptions in _frselect_ and the direct uses of _frdrawsel_ of _libframe_ in acme, sam and other clients; the assumptions being that the adjacent characters never overlap. Being much, much trickier to address, those fall outside of the scope of the proposed fix.

In #766 a similar fix was made to the macOS version of the program, and yrk-lab/fontsrv#4 is equivalent to the proposed fix.

Fixes #765 for X11 systems (Linux).

Tested on Raspbian (Debian) Trixie.

### DIFF
--- a/src/cmd/fontsrv/x11.c
+++ b/src/cmd/fontsrv/x11.c
@@ -199,13 +199,13 @@ mksubfont(XFont *xf, char *name, int lo, int hi, int size, int antialias)
 		for(r=0; r < bitmap->rows; r++)
 			memmove(base + r*mc->width*sizeof(u32int), bitmap->buffer + r*bitmap->pitch, bitmap->pitch);
 
-		memimagedraw(m, Rect(x, 0, x + advance, y), mc,
-			Pt(-face->glyph->bitmap_left, -(y - y0 - face->glyph->bitmap_top)),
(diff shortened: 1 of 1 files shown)


yrk-lab:
@dancrossnyc

PR #768, opened 2026-04-22
fix INSTALL when rc-related environment vars are set

rogpeppe:
If the variables FORCERCFORMK or MKSHELL
are set in the user's environment, the Plan 9 build will fail, so avoid that by unsetting them in the INSTALL script.

Patch courtesy of Jeff Sickel


yrk-lab (approved):
lgtm, thanks for the fix!

Still needs one of the maintainers to accept & merge.

dancrossnyc (approved):

PR #771, opened 2026-05-05
p: fix path to rc executable for ! command

dmitrivereshchagin:


dancrossnyc (approved):
Ok. Seems reasonable.

PR #774, opened 2026-05-31
dc: fix path to rc executable for ! command

dmitrivereshchagin:


dancrossnyc (approved):
I think this is fine as-is, though it points to an unfortunate impedance mismatch between Unix and Plan 9, and I offer a couple of alternatives you may want to consider, assuming `rc` is in the Unix `$PATH`:

I see that a fair bit of our code already uses `execlp`, which would solve this with no further modifications (save removing the `/bin/` prefix).

I also see that our version of `execl` is a wrapper around `exec`, which in turn wraps `execvp`; so an alternative may be to just remove `/bin/` from the path.

Anyway, I'm not saying that you need to change the change, but either of those (`execlp` in particular) would be smaller deltas.

dmitrivereshchagin:
It will be better with execlp. Thanks! Updated.

dancrossnyc (approved):

PR #779, opened 2026-06-25
factotum: do not guess the secstore address

0intro:
On Plan 9, factotum uses ndb to find whether the authentication domain has a secstore. There is no ndb here, so secdial guessed the address by dialing $auth or the host "secstore".

This blocks factotum's startup, and thus its 9P service, when there is no secstore: the auth server is not a secstore and stalls the read, and resolving "secstore" can hang.

Only dial a secstore explicitly set through $secstore.

### DIFF
--- a/src/cmd/auth/factotum/secstore.c
+++ b/src/cmd/auth/factotum/secstore.c
@@ -24,12 +24,12 @@ secdial(void)
 	char *p;
 
 	p = secstore;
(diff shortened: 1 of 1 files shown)


0intro:
CC @millerresearch It should improve reliability of our Go builders.

dancrossnyc (approved):

PR #780, opened 2026-06-25
factotum: do not fail the rpc read with a reply pending

0intro:
A conversation produces a reply with rpcrespond, which buffers it and resets the rpc op to RpcUnknown.

The conversation can produce the reply before the
matching read of /mnt/factotum/rpc arrives. The read then finds the op reset and responds "no rpc pending", even though a reply is buffered.

This race is timing dependent and rarely seen, but some clients trigger it often, breaking p9sk1
authentication through auth_proxy.

This change delivers the buffered reply in that case, instead of failing the read.


0intro:
CC @millerresearch It should improve reliability of our Go builders.

dancrossnyc (approved):

PR #781, opened 2026-06-28
devdraw: do not change window position when resizing on X11

dmitrivereshchagin:
Without this change, drawresizewindow(3) moves window to the upper-left corner of the screen.


dmitrivereshchagin:
Sample code to reproduce the issue:

```
// 9c foo.c && 9l -o foo foo.o && foo
#include <u.h>
#include <libc.h>
#include <draw.h>
#include <event.h>

void
eresized(int){}

void
main(int, char*[])
{
	initdraw(nil, nil, nil);
	einit(Emouse);
	for(;;){
		if(emouse().buttons)
			drawresizewindow(Rect(0, 0, 300, 300));
	}
}
```

